Issue #61, opened 2014-11-05
Add command line option to set theme


leif81:
Maybe I'm overlooking the obvious, but it appears the only way to change the theme yml is to modify the source code (https://github.com/asciidoctor/asciidoctor-pdf/blob/master/lib/asciidoctor-pdf/theme_loader.rb#L13) ? 

If this is the case, lets add a command line arg to set the theme.

PR #63, opened 2014-11-06
Add example for specifying theme file

leif81:
See issue #61 for my confusion that lead to adding this example here.


bfad:
This is no longer part of the documentation, and it should be re-added or the documentation should be updated with the correct way to do this.

mojavelinux:
Isn't it here?

https://github.com/asciidoctor/asciidoctor-pdf/blob/master/docs/theming-guide.adoc#applying-your-theme

If there's information missing from that section that you need, please let me know and I'll add it.

bfad:
I missed the "and applying" part when reading the README. That has everything I need, thanks!

mojavelinux:
Great! I'm glad to hear that. Thanks for replying.

Issue #58, opened 2014-10-20
Fallback fonts support


chloerei:
Seems like fallback fonts settings in theme.yml not working yet, please implement the feature in the future, without fallback fonts I can not render an document write in CJK characters.

The project is great! thanks for all contributors.

mojavelinux:
Resolved by 11bcf455f39429706ee5890eb54886ee6a01bced.

mojavelinux:
Forgot to close. Thanks @chloerei!

PR #77, opened 2015-01-01
enable font_fallbacks support

chloerei:
Let editor can use font fallbacks settings in themes.


chloerei:
Relate to https://github.com/asciidoctor/asciidoctor-pdf/issues/58

mojavelinux:
The fallback_fonts feature was previously disabled because it was breaking global styling. This was fixed in Prawn 1.2.1 as described in the following section in the release notes:

https://github.com/prawnpdf/prawn/wiki/CHANGELOG#fallback-fonts-no-longer-break-global-font-styling

It's safe to use this feature. I just confirmed it works as expected.

mojavelinux:
Thanks for fixing this, @chloerei!

mojavelinux:
I'm going to reenable the fallback_fonts in the default asciidoctor theme to use the already provided M+ 1p multilingual font. We know it doesn't cover all CJK characters, but at least it catches some for now. We'll address more complete support as part of #82.

Issue #65, opened 2014-11-14
Be able to set custom page sizes for pdf 


IbnSaeed:
Hello

Would it be possible to add the ability to set custom page sizes such as A5 or using any custom dimensions such as 134mm x 233mm etc.

Specifically targeting Printed Books

IbnSaeed:
Was this released in 1.5 ?

mojavelinux:
Not yet. This is still scheduled to be completed. I'll mark it as a hackfest item as it's a nice self-contained change.

mojavelinux:
I decided to go with the attribute name `pdf-page-size`. Note that this attribute now supports setting the width and height using dimensions with units, including in, cm, mm and pt.

For example:

```
:pdf-page-size: [8.25in, 11.69in]
```

Units can also be used in the theme file.

mojavelinux:
Let me know if this looks good and I'll merge.

mojavelinux:
I've added support for the syntax `width x height` (e.g., `8.5in x 11in`). I also switched to a manual and more strict parser instead of delegating to YAML.

mayeulk:
Hi,
I had difficulties having this work (especially since https://github.com/asciidoctor/asciidoctor-pdf/issues/1518 says "widht x heigh" is OK) **with the command line**. So this might help others.
This works for me:

     asciidoctor-pdf --attribute=pdf-page-size="[14cm, 14cm]" myfile.asciidoc

The following DOES NOT worked:
`asciidoctor-pdf  attribute=pdf-page-size="[14cm x 14cm]" action-bookmark.asciidoc`

Config:
Asciidoctor PDF 2.3.19 using Asciidoctor 2.0.23 [https://asciidoctor.org]
Runtime Environment (ruby 3.3.0 (2023-12-25 revision 5124f9ac75) [x86_64-linux]) (lc:UTF-8 fs:UTF-8 in:UTF-8 ex:UTF-8)

mojavelinux:
Please ask questions in the project chat at https://chat.asciidoctor.org. The issue tracker is not a support forum, especially closed issues.

PR #84, opened 2015-01-10
theme: Allow custom page size setting

otavio:
To allow a custom page size to be used, we need to check if the
argument passed in the theme is a String, for which we ought to upcase
to use in Prawn converted, or keep the custom value to be passed over
it.

This has been tested using a custom theme with:

,----
| ...
| page:
|   ...
|   size: [508, 635]
|   ...
| ...
`----

Signed-off-by: Otavio Salvador <otavio@ossystems.com.br>

### DIFF
--- a/lib/asciidoctor-pdf/converter.rb
+++ b/lib/asciidoctor-pdf/converter.rb
@@ -165,9 +165,12 @@ class Converter < ::Prawn::Document
       info: (build_pdf_info doc),
       margin: (theme.page_margin || 36),
       page_layout: (theme.page_layout || :portrait).to_sym,
-      page_size: (theme.page_size || 'LETTER').upcase,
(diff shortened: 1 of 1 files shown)


mojavelinux:
Nice improvement!

I have one suggestion. Could you write the assignment as follows?

```ruby
pdf_opts[:page_size] = pdf_opts[:page_size].upcase if ::String === pdf_opts[:page_size]
```

In Ruby, `===` is equivalent to `is_a?` in this arrangement, but slightly faster. I also avoided the use of string modification because it's best to avoid that practice.

mojavelinux:
This partially addresses #65. I say partially because it must be set in the theme file rather than the document, so we'll still need a change to allow it to be set in by the document.

mojavelinux:
This looks great @otavio! Thanks!

PR #109, opened 2015-05-08
Fixed #104, table header background color

### LINKED ISSUE #104: Table header
Hello. 
Is there any way to style the table header seperatly? I have can se that the rows kan be styled with different background colors on odd an even row numbers, but I can't find out how to style the tabel header.
With Regards
Ole

theimdal:
Also added the possibility to specify the background color for even table rows.

### DIFF
--- a/data/themes/default-theme.yml
+++ b/data/themes/default-theme.yml
@@ -176,8 +176,8 @@ table:
   head_font_color: 000000
   head_background_color: FFFFFF
   background_color: transparent
-  background_color_row_even: FFFFFF
-  background_color_row_odd: FFFFFF
+  even_row_background_color: FFFFFF
+  odd_row_background_color: FFFFFF
   border_color: dddddd
   border_width: $base_border_width
   # HACK accounting for line-height
--- a/data/themes/default-theme.yml
+++ b/data/themes/default-theme.yml
@@ -173,8 +173,11 @@ description_list:
 outline_list:
   indent: $horizontal_rhythm * 1.25
 table:
+  head_font_color: 000000
+  head_background_color: FFFFFF
   background_color: transparent
-  background_color_alt: f9f9f9
+  background_color_row_even: FFFFFF
+  background_color_row_odd: FFFFFF
   border_color: dddddd
   border_width: $base_border_width
   # HACK accounting for line-height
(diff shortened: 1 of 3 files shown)


mojavelinux:
Thanks for contributing this fix! I've made a few comments inline. If you could address those, then I'll merge.

data/themes/default-theme.yml
```diff
@@ -173,8 +173,11 @@ description_list:
 outline_list:
   indent: $horizontal_rhythm * 1.25
 table:
+  head_font_color: 000000
+  head_background_color: FFFFFF
   background_color: transparent
-  background_color_alt: f9f9f9
+  background_color_row_even: FFFFFF
+  background_color_row_odd: FFFFFF
```
mojavelinux: Let's use these names instead:

```
even_row_background_color: FFFFFF
odd_row_background_color: FFFFFF
```

That makes it consistent with `head_background_color` and so forth.

lib/asciidoctor-pdf/converter.rb
```diff
@@ -153,6 +153,7 @@ def init_pdf doc
     register_fonts theme.font_catalog, (doc.attr 'scripts', 'latin'), (doc.attr 'pdf-fontsdir', ThemeLoader::FontsDir)
     @theme = theme
     @font_color = theme.base_font_color
+    @background_color = theme.page_background_color
```
mojavelinux: I don't think we need to make `@background_color` an instance variable. I'm only doing that for `@font_color` because it saves a lot of typing.

mojavelinux:
Could you update the data/themes/asciidoctor-theme.yml file as well? I think the following values will do:

```
table:
  background_color: transparent
  head_background_color: F9F9F9
  head_font_color: $base_font_color
  even_row_background_color: F9F9F9
  odd_row_background_color: transparent
  ...
```

mojavelinux:
Note that we may need to check for the value of "transparent" and not pass it through as it might confuse Prawn. Prawn doesn't handle the value of "transparent" internally.

theimdal:
Hi.

I have implemented the changes you asked for, but I have not done anything in regards to the "transparent" issue.

mojavelinux:
> I have implemented the changes you asked for

Super! Thanks!

> but I have not done anything in regards to the "transparent" issue.

I'll take care of that. I'm not quite sure what needs to be done, so I plan to play around with it. I'll come back once I know more.

mojavelinux:
Nice work! I made some changes to handle transparent values, but still in line with your changes. Note that I did need to reverse the even and odd so that it applied to the correct lines (odd and even is based on body rows).

theimdal:
Refinements are always welcome :)

mojavelinux:
\o/

PR #115, opened 2015-05-13
resolves #65 support setting page size via attribute

### LINKED ISSUE #65: Be able to set custom page sizes for pdf

mojavelinux:
- support setting page size via pdf-page-size attribute
- add support for units in page size
- validate named page size
- document page size options in default theme file

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -6,6 +6,8 @@ Dan Allen <https://github.com/mojavelinux[@mojavelinux]>; Sarah White <https://g
 :idprefix:
 :idseparator: -
 ifdef::env-github[:relfileprefix: /blob/master]
+//:pdf-page-size: [8.25in, 11.69in]
+//:pdf-page-size: A4
 // Aliases:
 :project-name: Asciidoctor PDF
 :project-handle: asciidoctor-pdf
--- a/data/themes/default-theme.yml
+++ b/data/themes/default-theme.yml
@@ -33,6 +33,7 @@ page:
   layout: portrait
   # multiply inches by 72 to get pt values
   margin: [0.5 * 72, 0.67 * 72, 0.67 * 72, 0.67 * 72]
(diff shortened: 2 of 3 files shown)


mojavelinux:
@IbnSaeed how does this look?

IbnSaeed:
Excellent. Thanks alot for adding this feature.

mojavelinux:
Sure thing!

One thing I wasn't sure about is whether to make the layout (i.e., orientation) a document-level attribute too. Technically, switching the layout between portrait and landscape can be accomplished simply by flipping the values in the pdf-page-size. In fact, that's all setting the layout to landscape does (in Prawn, at least).

I was also wondering if we should support an alternate syntax for the value using `x` as a delimiter:

```
:pdf-page-size: 8.25in x 11.69in
```

Do you think that's worth doing or is the array-like syntax sufficient?

IbnSaeed:
I use LaTeX with memoir class for book layouts and this is how its defined in memoir|

`
\setstocksize{9in}{4in}
`
The array syntax would be sufficient `pdf-page-size: [8.25in, 11.69in]`, is fine by me.

If its not a hassle, you could add a LaTeX style syntax as well .

mojavelinux:
I think I'm going to move forward with the syntax I proposed. It just seems very natural and not difficult to parse.

> If its not a hassle, you could add a LaTeX pdf-page-size{9in}{4in} style syntax as well .

I'm not considering that since the goal of AsciiDoc is not to mimic LaTeX, only to provide a suitable replacement. It's important to adhere to the AsciiDoc style for our syntax.

mojavelinux:
The `width x height` syntax is now implemented (will be available in 1.5.0.alpha.8).

PR #122, opened 2015-05-14
Fix typo in prawn-icon include in gemspec

DavidGamba:


mojavelinux:
Oops, looks like a merge error. Thanks for catching!!

PR #124, opened 2015-05-17
resolves #87 properly resolve image target as system path

### LINKED ISSUE #87: Images are not being correctly inserted
Images which are in the images directory are not included in the final rendered document.
During render the following message is shown:
```
asciidoctor: WARNING: could not embed image; images/${LOCAL_PATH}/asciidoctor-maven-examples/asciidoctor-pdf-example/src/docs/asciidoc/images/sunset.jpg n
ot found
```

Seems like the path is incorrectly appended at the beginning of the full path.
As an example you can test it with `asciidoctor-pdf-example` in [asciidoctor-maven-examples](https://github.com/asciidoctor/asciidoctor-maven-examples).

This has been tested setting the `images` parameter in the pom.xml to other values and the behaviour is the same, the parameter's value is appended at the beginning.

Related to asciidoctor/asciidoctor-maven-examples#11

mojavelinux:
It's not pretty yet, but I think this resolves images in a way consistent with the html5 backend when data-uri is set.

### DIFF
--- a/lib/asciidoctor-pdf/converter.rb
+++ b/lib/asciidoctor-pdf/converter.rb
@@ -634,8 +634,10 @@ class Converter < ::Prawn::Document
     #move_down @theme.block_margin_top unless at_page_top?
     theme_margin :block, :top
     target = node.attr 'target'
-    #if target.end_with? '.pdf'
-    #  import_page target
+    # FIXME API should provide a cleaner and clearer way to resolve image to system path
+    image_path = ::File.expand_path node.normalize_system_path(target, (node.document.attr 'imagesdir'))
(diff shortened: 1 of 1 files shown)


mojavelinux:
I think this is good to go. It's prettier now...with the exception of the temporary file handling. But I'll let that simmer and find a way to improve it later.

Here's what we have now:

* warn if image is gif
* proper resolution of image type based on file extension
* proper resolution of image path (matching how the html5 converter works when resolving an image to embed)
* resolve image from URI if allow-uri-read is set (no safe mode checking atm)
* happier users :)

PR #126, opened 2015-05-18
resolves #110 fix xrefs to non-section content

### LINKED ISSUE #110: Cross references don't work when target is a block
Cross references don't work when the target is a block. To reproduce, assign an id to a block.

```
.Pretty picture
[#pretty-picture]
image::pretty-picture.png[Pretty Picture,600,400]
```

Then link to that block.

```
See the <<pretty-picture>>.
```

A link is created, but when you click on it, it does not work.

mojavelinux:
- introduce add_dest_for_block for adding xref targets (i.e, destinations)
- call add_dest_for_block in each block handler
- add support in parser for <a name> to set inline destinations
- handle render_behind callback to add inline destinations
- enable inline ref and bibref targets


mojavelinux:
With this pull request, all xrefs should now be working.

mojavelinux:
This depends on #127 being merged.

PR #128, opened 2015-05-19
resolves #31 process callouts in source blocks when using CodeRay

### LINKED ISSUE #31: Callout support
`asciidoctor-pdf` does not seem to support callouts. Is there a way to enable this?

This may or may not be related to https://github.com/asciidoctor/asciidoctor-pdf/issues/15, I'm not sure.

mojavelinux:
- add method for extracting conums from source string
- add method for restoring conums to source fragments
- add method for resolve conum glyph
- minor cleanups to CodeRay encoder


mojavelinux:
I'm confident we now have callout numbers working properly in all listing and literal blocks, including when the source is highlighted using CodeRay or Pygments and when it's not.

Note that the callout numbers are open circles, which is different from how they look in the HTML output. This is because only the open circles are consecutive in unicode. We need to make this configurable and also lookup the numbers from a map instead of our current hack.

PR #137, opened 2015-05-21
resolves #136 unencode character entities in document title

### LINKED ISSUE #136: Unescape entities in the document title in PDF info and outline
Unescape entities in the document title before putting it into the PDF info and outline. Both places require literal, not interpreted text.

For example, consider the following title:

```
Source Code & Syntax Highlighting
```

Currently, a `&amp;` is showing instead of `&` in the document title and outline.

mojavelinux:
- unencode char entities in document title in PDF info, outline & pdfmarks
- encode PDF info as hex if content cannot be converted to ASCII-8BIT
- promote regexps used in sanitize method as constants
- update example to include entity and UTF-8 in PDF info
- minor code cleanups


mojavelinux:
Note that this pull request also fixes encoding errors in the PDF info when the string contains a non-ASCII character. You should no longer see the following warning:

```
warning: regexp match /.../n against to UTF-8 string
```

This was happening because we were trying to put a UTF-8 string into a PDF LiteralString which only accepts ASCII-8BIT encoded strings. It's necessary to use a hex-encoded value if the string contains non-ASCII characters.

mojavelinux:
I organized the code a bit more, making use of modules to avoid redundant code and unnecessary passing of references. I think it's good to go now.

PR #141, opened 2015-05-22
resolves #138 support text-transform property for headings

### LINKED ISSUE #138: Add support for text-transform property in theme
Add support for text-transform property in the theme, particularly for headings. Be careful not to uppercase named character entities (i.e., `&amp;`)

```
heading:
  text_transform: uppercase
```

Support the following values:

* uppercase
* lowercase
* capitalize

Ideally, the theme value should be handled anywhere we use `theme_font`. However, Prawn doesn't provide a low-level transform for this, so we actually have to modify the string.

mojavelinux:


mojavelinux:
The initial pull request sets the `@text_transform` instance variable on the class whenever the `theme_font` DSL method is used. However, it currently only applies the `@text_transform` to headings (with some edge cases).

mojavelinux:
Add support for cells in table head as well.

mojavelinux:
The entry in the table of contents should match the text_transform of the heading to which it matches.

mojavelinux:
> The entry in the table of contents should match the text_transform of the heading to which it matches.

I decided to make this a theme setting instead. You could link them as follows:

```
toc:
  text_transform_h2: $heading_text_transform_h2
```

You can also set the text transform for the whole toc.

```
toc:
  text_transform: uppercase
```

PR #145, opened 2015-05-24
resolves #134 allow page background image to be specified

### LINKED ISSUE #134: Allow a background image to be specified for the title page
Allow a background image to be specified for the title page. This should likely go in the theme, but since it can be closely associated with the content of the document, we can also permit it to be specified using a document attribute.

The image path should be resolved relative to the imagesdir if specified in the document and perhaps the stylesdir if specified in the theme.

For the document attribute, I propose the following name:

```
:title-page-background-image: <path>
```

For the theme, it should be specified under the key:

```
title_page:
  background_image: <path>
```

mojavelinux:
- allow background image to be specified for all pages
- allow background image to be specified for title page
- resolve background image relative to pdf-stylesdir (theme dir)
- allow background color to be specified for title page
- add fill_absolute_bounds convenience method

### DIFF
--- a/lib/asciidoctor-pdf/converter.rb
+++ b/lib/asciidoctor-pdf/converter.rb
@@ -95,12 +95,14 @@ class Converter < ::Prawn::Document
     # data-uri doesn't apply to PDF, so explicitly disable (is there a better place?)
     doc.attributes.delete 'data-uri'
 
-    # TODO implement page_background_image as alternative and/or page_watermark_image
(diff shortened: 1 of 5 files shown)


mojavelinux:
The only thing left to support is a title page background per document, read from a document attribute.

mojavelinux:
Per-document title page background attribute implemented. If the image is specified in the document, the path is resolved relative to the imagesdir. If the image is specified in the theme, it is resolved relative to the pdf-stylesdir, if specified, or built-in theme directory.

PR #174, opened 2015-06-05
resolves #97 add support for inline icon macro

### LINKED ISSUE #97: Icon fonts missing from PDF
If `:icons: font` is applied to the document then icons go missing in the final PDF. Asciidoctor outputs the following warning

    asciidoctor: WARNING: conversion missing in backend pdf for inline_image

gradle 2.3
asciidoctor 1.5.2
asciidoctorj-pdf 1.5.0-alpha.6

mojavelinux:
- add support for inline icon macro
- allow font set to be changed globally (icon-set=<icon-set> attribute)
- allow font set to be changed per icon (set=<icon-set> attribute or append @<icon-set> to target)

### DIFF
--- a/lib/asciidoctor-pdf/converter.rb
+++ b/lib/asciidoctor-pdf/converter.rb
@@ -55,6 +55,7 @@ class Converter < ::Prawn::Document
     circle: (unicode_char 0x25e6),
     square: (unicode_char 0x25aa)
   }
+  IconSets = ['fa', 'fi', 'octicon', 'pf']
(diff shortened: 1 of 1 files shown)


mojavelinux:
Full example:

```
Available now from icon:amazon@pf[].
```

mojavelinux:
In addition to the sizes supported in the HTML backend (lg, 1x, 2x, etc), you can enter any relative value in the size attribute (e.g., 1.5em, 150%, etc).

PR #178, opened 2015-06-09
resolves #9 support inline images

### LINKED ISSUE #9: Add support for inline images
Add support for the inline image macro. This requires adding a convenience method for placing an image in the flow of text. It would be nice to have direct support for this in Prawn, though we can get away with using a method in Asciidoctor PDF until it becomes available upstream.

mojavelinux:
- arrange and embed inline images (currently don't affecting line height)
- support remote inline images
- use alt text if image can't be found or read
- move text formatter code under Asciidoctor::Pdf::FormattedText namespace
- only match formatted text that the converter creates
- optimize how Prawn extension code is loaded
- code cleanups


mojavelinux:
I'm holding on this one because I want to see if I can allow the line height to be adjusted if the image doesn't fit in the current line height.

mojavelinux:
I found a way to adjust the line height. If the image height is more than 1.5x time the font height (roughly the line height with some fudge space), then I mess with the ascender and descender values to push the height of the whole line up so that it can accommodate the image. In this case, the bottom of the image will be aligned with the bottom of the text (actually, slightly below). At this point, Prawn just doesn't provide enough control to align the image any other way.

mojavelinux:
If the image height doesn't exceed the height of the line (roughly), then the image is centered in the line of adjacent text.

PR #184, opened 2015-06-11
resolves #181 enable checklists

### LINKED ISSUE #181: Enable Checklist In pdf Output
asciidoctor-pdf   doesn't allow  to have checklist in the ouput pdf .  Would be nice to have this functionality . 

Cheers

mojavelinux:
* add ballot box glyphs (0x2610 and 0x2611) to notoserif-regular font
* use ballot box glyphs as marker for checklists


mojavelinux:
Note that we are copying the ballot box glyphs from Font Awesome into NotoSerif. This allows you to still use a custom font that provides it's own ballot box glyphs.

In the future, we might consider using the Font Awesome font regardless of the font being used for text.

PR #186, opened 2015-06-12
resolves #185 autofit font size in listing & literal blocks

### LINKED ISSUE #185: Making a listing block text smaller
I have some output that I want to make smaller so that it doesn't wrap and looks more like my gedit screen. I tried using the [small] attribute, and it does make it smaller but leaves [small] visible in the output and still wraps. I'm pretty sure I'm just not doing it right.

Question
1) Can I make listing block text smaller than what [small] does?
2) How do I get rid it showing [small] # # 

What my gedit screen looks like with asciidoc

selection_212

Output looks like:
selection_213

mojavelinux:
- autofit (reduce font size) to prevent wrapping in listing & literal blocks
- prevent font size from being reduced more than *_font_size_min
- activate autofit if autofit option is specified on block
- activate autofit if autofit-option attribute is specified on document
- add autofit example to chronicles.adoc
- add additional helper methods to core_ext


mojavelinux:
Note that this fix has some gaps. If the longest line has a callout number or the longest line uses inline formatting, the adjustment will overcompensate, making the text smaller than it has to be. This is a tricky...and expensive...calculation to get precise.

mojavelinux:
I think for now, this is a good start.

mojavelinux:
Is `source-autofit` too specific? Should it be `option-autofit` instead?

mojavelinux:
I think this is good to go now.

mojavelinux:
@rlopez133 does this work for you now?

rlopez133:
When I try to build my pdf, getting this error:

````
/root/asciidoctor-pdf/lib/asciidoctor-pdf/converter.rb:9:in `require_relative': cannot load such file -- /root/asciidoctor-pdf/lib/asciidoctor-pdf/pdf_core_ext (LoadError)
	from /root/asciidoctor-pdf/lib/asciidoctor-pdf/converter.rb:9:in `<top (required)>'
	from /root/asciidoctor-pdf/lib/asciidoctor-pdf.rb:3:in `require_relative'
	from /root/asciidoctor-pdf/lib/asciidoctor-pdf.rb:3:in `<top (required)>'
	from /root/asciidoctor-pdf/bin/asciidoctor-pdf:3:in `require_relative'
	from /root/asciidoctor-pdf/bin/asciidoctor-pdf:3:in `<main>'
````

mojavelinux:
Are you sure you are running it on the adoc file? I'll double check, but that seems to be the only way we'd see a DOCTYPE in the output.

rlopez133:
Ya running it on the same doc file. I reverted the changes to get adoc-pdf working again. Not sure why it was not happy. I manually copied the files since it wasn't avail via the git pull. This could prob be part of my prob maybe I missed something....

mojavelinux:
Gotcha. I'll verify. If it looks clean from my end, I'll merge and then you can test again.

PR #197, opened 2015-06-15
resolves #196 add support for border style

### LINKED ISSUE #196: Allow border style to be specified (for running content and thematic break)
All the border style to be specified, starting with the lines used in running content and thematic breaks.

The following values should be supported:

* solid (default)
* double
* dashed
* dotted

double should leave a gap between the lines that is 3 times the width of the line.

dashed should set the dash length to 3 times the width of the line

dotted should set the dash length equal to the width of the line.

mojavelinux:
- add support for line_style to stroke_horizontal_rule method
- allow border style to be set for running header/footer and thematic break
- document new keys
- lowercase all color values and prefix with # in theming guide

### DIFF
--- a/docs/theming-guide.adoc
+++ b/docs/theming-guide.adoc
@@ -662,14 +662,15 @@ TBW
 
 |page_background_color
 |<<colors,color>>
-|background_color: ffffff
+|background_color: #ffffff
 
 |page_background_image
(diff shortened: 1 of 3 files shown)


mojavelinux:
cc: @rlopez133

PR #203, opened 2015-06-17
resolves #190 dash border where literal/listing block is broken across pages

### LINKED ISSUE #190: Make a visible continuation hint for listing that spans more than one page
Make a visible continuation hint at the top and bottom of a listing block where it is separated by a page break. Otherwise, it appears as though there are two different listing blocks.

This should only be done if the listing block has a border. If it does, draw a dashed border to indicate that the listing has been separated by a page break.

mojavelinux:


mojavelinux:
cc: @rlopez133 @robertpanzer

PR #204, opened 2015-06-18
resolves #68 upgrade Prawn

### LINKED ISSUE #68: undefined method `on_page_create' for #<Asciidoctor::Pdf::Converter:0x007ff89307f6c8>
The following error appears when trying to convert a document:

undefined method `on_page_create' for #<Asciidoctor::Pdf::Converter:0x007ff89307f6c8>

This error seems new to asciidoctor-pdf-1.5.0.alpha.6 release.

mojavelinux:
- set Prawn version to >= 1.3.0 and < 3.0.0
- convince Bundler to install Prawn 1.3.0 if Ruby version < 2.0.0
- resolve missing on_page_create method
  - add check for on_page_create method since its status is questionable

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,6 @@
 source 'https://rubygems.org'
 
+gem 'prawn', '1.3.0' if (Gem::Version.new RUBY_VERSION) < (Gem::Version.new '2.0.0')
+
 # Look in asciidoctor-pdf.gemspec for runtime and development dependencies
 gemspec
--- a/README.adoc
+++ b/README.adoc
@@ -39,7 +39,7 @@ See the milestone v1.5.0 in the {uri-project-issues}[issue tracker] for details.
 
 == Prawn, the majestic PDF generator
 
-{uri-project}[{project-name}] is made possible by an amazing Ruby Gem named Prawn.
+{uri-project}[{project-name}] is made possible by an amazing Ruby gem named Prawn.
 And what a gem it is!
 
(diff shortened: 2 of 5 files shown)


mojavelinux:
Note that this is going to cause some stumbling blocks for AsciidoctorJ. The build will need to use JRuby 9.0.0.0.rc1 to prepare gems in the Gradle build since Prawn requires Ruby > 2.0.0 at installation time. However, once Prawn is bundled, it will work find with JRuby 1.7 in either mode.

We'll need the following in the asciidoctorj-pdf/build.gradle file:

```
jruby {
  execVersion = '9.0.0.0.rc1'
}
```

(or if you can figure out how to enable 2.0 mode when running jrubyPrepareGems)

We'll also need to update the version of the Prawn and TTFunk gems in the root build.gradle file:

```
prawnGemVersion = '2.0.1'
ttfunkGemVersion = '1.4.0'
```

(though, perhaps it's no longer necessary to specify these versions explicitly).

mojavelinux:
I found a way to allow Prawn 1.3.0 to be used. Thus, it should be possible for AsciidoctorJ to specify the Prawn version as 1.3.0 and avoid any Ruby 2.0 hacks for now. That should ease the process for releasing alpha8.

mojavelinux:
@jessedoyle What do you think? Does this look good?

jessedoyle:
@mojavelinux: Looks good to me! Thanks for getting this update in!

mojavelinux:
Thanks for taking a look at it Jesse!

PR #225, opened 2015-06-24
small fix for simple docs without sections and doctitle

abatalev:


mojavelinux:
Resolves #229.

Thanks @abatalev!

PR #226, opened 2015-06-24
resolves #224 add support for line numbers in source listings

### LINKED ISSUE #224: Code line number are not shown in generated pdf
The linenums is a great feature! This works fine in the asciidoctor generated HTML:

[source, javascript, linenums]

But asciidoctor-pdf seems to ignore linenums and no line numbers are shown in the generated PDF.

mojavelinux:


mojavelinux:
This pull request is currently a very crude implementation. It needs to be reworked / optimized before merging.

mojavelinux:
Line numbers are supported when using Rouge as the syntax highlighter. Since the plan is to drop support for CodeRay, I think this pull request should be rewritten to only add line numbers for Pygments.

mojavelinux:
This has been rewritten to only support Pygments.

mojavelinux:
I think lineno_font_color should be changed to code_linenum_font_color (or code_lineno_font_color).

mojavelinux:
Updated (to code_linenum_font_color) (since the conum category is singular).

PR #244, opened 2015-06-27
resolves #236 expand tabs to spaces

### LINKED ISSUE #236: Syntax code blocks lose indenting but keep colours in PDF
Code that is rendered correctly in the [HTML output ](https://rawgit.com/openEHR/spec-publish-asciidoc/master/docs/AOM2/AOM2.html#_conformance_semantics) is coloured correctly (indicating that Pygments styling is occurring), but the indenting is lost in the PDF:

image

mojavelinux:


mojavelinux:
We may just put this into Asciidoctor 1.5.3 and assume that it's handled once we enter the converter.

mojavelinux:
I wonder if tabsize should be a document-level attribute or allowed on the block itself. Usually, this is something you make consistent throughout the document, though I can see cases where the tabsize should different the content (perhaps columnar data) comes from a place that assumes a specific tabsize.

mojavelinux:
I think this is ready to go now, though it will still have to see how things play out in core and may have to adjust the implementation accordingly.

mojavelinux:
Core now handles tab expansion, so this logic only kicks in if tabs are not replaced (since they must be replaced or else they won't render properly).

PR #278, opened 2015-07-24
Control toc position with toc-start attribute

### LINKED ISSUE #233: Can't control PDF TOC output location
I am unable to prevent the TOC coming out as the very first set of pages in the PDF. I want them in a controllable location later in the document (after Amendments table, Acknowledgements).

I tried `:toc: macro` at the top, and `toc::[]` at the desired location, but this has no effect. Ideally I need a solution that also enables the same source to generate an HTML output with TOC on the left. I don't mind invoking asciidoctor with different command lines if need be to achieve this.

UPDATE: I discovered that if I put :toc: macro below the title heading (first = xxxx line), that no TOC at all comes out even with the toc::[] later on. No idea what is going on now. Why does it matter where the :toc:  line goes?

hmflash:
Allow toc placement to be controlled with 'toc-start' attribute.

This fixes #233 

If attribute is omitted, toc is placed immediately following the cover page (the same as it is now).

Changing the value to 3 allows for the an abstract or introduction or copyright notice to be placed prior to the table of contents.

An example adoc that puts the toc after the abstract would look like:

```
= My document title
:doctype: book
:toc:
:toc-start: 3

[abstract]
The abstract goes here

== First document section

First document section content
```

### DIFF
--- a/lib/asciidoctor-pdf/converter.rb
+++ b/lib/asciidoctor-pdf/converter.rb
@@ -145,19 +145,6 @@ class Converter < ::Prawn::Document
 
     start_new_page
 
-    toc_start_page_num = page_number
-    num_toc_levels = (doc.attr 'toclevels', 2).to_i
-    if (include_toc = doc.attr? 'toc')
-      toc_page_nums = ()
-      dry_run do
-        toc_page_nums = layout_toc doc, num_toc_levels, 1
-      end
-      # NOTE reserve pages for the toc
-      toc_page_nums.each do
-        start_new_page
-      end
-    end
-
     num_front_matter_pages = page_number - 1
     font @theme.base_font_family, size: @theme.base_font_size
     convert_content_for_block doc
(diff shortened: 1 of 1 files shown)


mojavelinux:
Thanks for submitting this PR!

I'll need to review this very carefully (testing against a bunch of different edge cases) since placing the toc into the document is a very tricky ordeal atm and the logic around it is...a little fragile to say the least. However, it's a great opportunity to get it cleaned up.

hmflash:
I just found and fixed a bug during more of my own testing.  If the toc attribute was omitted, we still need the number of toc levels for the outline.

flaub:
Any way this can get merged?

mojavelinux:
Until I can review this thoroughly, I can't be positive this doesn't break other use cases. The toc situation is very tricky because it has to work forwards and backwards.

Having said that, it is possible to use this code in your own project. You can see the InfoQ mini book as an example of how to override the code in Asciidoctor PDF to perform customizations similar to this one.

See:

* https://github.com/mraible/infoq-mini-book/blob/master/src/main/ruby/asciidoctor-pdf-extensions.rb
* https://github.com/mraible/infoq-mini-book/blob/master/generate-pdf.sh#L28

mojavelinux:
As a first step, the toc is now inserted between the document title and the body when the doctype is not article. More important, the groundwork is now in place to allow the toc to be moved around.

I've opened a new issue that proposes allowing the use of the toc macro to move the toc around. See #1030. As long as this feature can with tests, and not break an existing tests, then it can be merged. If the toc macro does not work, and we still need a special start attribute, let's discuss it in #1030.

mojavelinux:
After reviewing this, I don't think this is the right approach. It assumes you already know where content will fall when the content is rendered...thus overly coupled with the output. #1030 is a better approach. We may still be able to reuse this logic. But the start page of the toc will be implied by the location of the toc macro in the source document.

mojavelinux:
I finally figured out how to support the toc macro in Asciidoctor PDF. That's the more consistent and portable way to solve this problem. The implementation can be found in PR #1318.

PR #332, opened 2015-10-11
resolves #121. Admonition icon customization by theme.

### LINKED ISSUE #121: Allow font-based icon keys and colors to be set by the theme
Allow the font-based icons that are used when the document attribute `icons` is set to `font` to be customized using the theme. Currently, the icon data is hard-coded as a constant.

jessedoyle:
This patch allows for admonition icon customization via the pdf theme.

Here's an example of the theme syntax:

```yaml
admonition:
  icon:
    tip:
      key: fa-arrows
      color: 0099FF
      size: 14
```

Supported icon keys currently are: `key`, `color` and `size`. If any of these are not specified, they will remain as the Asciidoctor defaults.

I haven't tested the patch to ensure that I haven't broken something, so it's probably best to double-check that everything works okay!

### DIFF
--- a/lib/asciidoctor-pdf/converter.rb
+++ b/lib/asciidoctor-pdf/converter.rb
@@ -1782,10 +1782,14 @@ class Converter < ::Prawn::Document
   end
 
   def admonition_icon_data key
-    @theme.admonition_icons[key].tap do |data|
-      AdmonitionIcons[key].each do |k, v|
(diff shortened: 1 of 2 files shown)


mojavelinux:
Excellent work @jessedoyle! I'm eager to try it out!

mojavelinux:
This is fantastic! Such a great addition!

PR #335, opened 2015-10-13
update prawn-icon to 1.0.0

jessedoyle:
Update `Prawn/Icon` from `v0.6.4` to `1.0.0`. 

The only (potentially) breaking changes involve the use of a small number of icons removed from the `octicon` font. See the  [changelog](https://github.com/jessedoyle/prawn-icon/blob/master/CHANGELOG.md).


mojavelinux:
> The only (potentially) breaking changes involve the use of a small number of icons removed from the octicon font. 

Given how new the icon integration is in Asciidoctor PDF (and that Asciidoctor PDF is still alpha), I'm not concerned about these changes at all.

mojavelinux:
Thanks!

PR #336, opened 2015-10-14
Add icon rendering documentation with icon sets

anandchida:
Most of the icon enabling instructions is copied from issue #97. Also moved the important tip about sourcecode highlighting within the coderay section


mojavelinux:
Thanks a lot for adding this documentation, @anandchida!

PR #338, opened 2015-10-18
Add option to control generation of timestamps.

bk2204:
This is the corresponding change related to asciidoctor/asciidoctor#1453.

### DIFF
--- a/lib/asciidoctor-pdf/converter.rb
+++ b/lib/asciidoctor-pdf/converter.rb
@@ -294,7 +294,7 @@ class Converter < ::Prawn::Document
     info[:Creator] = str2pdfval %(Asciidoctor PDF #{::Asciidoctor::Pdf::VERSION}, based on Prawn #{::Prawn::VERSION})
     info[:Producer] ||= (info[:Author] || info[:Creator])
     # FIXME use docdate attribute
-    info[:ModDate] = info[:CreationDate] = ::Time.now
+    info[:ModDate] = info[:CreationDate] = ::Time.now unless doc.attr? 'reproducible'
     info
   end
 


lib/asciidoctor-pdf/converter.rb
```diff
@@ -296,7 +296,7 @@ def build_pdf_info doc
     info[:Creator] = str2pdfval %(Asciidoctor PDF #{::Asciidoctor::Pdf::VERSION}, based on Prawn #{::Prawn::VERSION})
     info[:Producer] ||= (info[:Author] || info[:Creator])
     # FIXME use docdate attribute
-    info[:ModDate] = info[:CreationDate] = ::Time.now
+    info[:ModDate] = info[:CreationDate] = ::Time.now if doc.attr 'timestamps'
```
mojavelinux: Could you write this as:

```ruby
if doc.attr? 'timestamps'
```

Notice the question mark.

mojavelinux:
We'll wait on the merge in core before merging this one.

lib/asciidoctor-pdf/converter.rb
bk2204: I wasn't clear if doing "timestamps!" would result in the value being absent altogether. If so, that's a fine solution and I'm happy to make the change. Otherwise, if it's still present but false, I think that will break functionality.

mojavelinux:
Now that the corresponding issue is merged in core, can you update this PR to align with the approach we took there (i.e., `reproducible` flag).

bk2204:
Sure.  I'll pick that up tomorrow, now that I'm back in Houston.

mojavelinux:
:+1: Thanks!

bk2204:
Updated.

mojavelinux:
Thanks!

PR #375, opened 2015-12-16
Set default page size to A4

### LINKED ISSUE #319: Default pdf-page-size
Hello,

the current default `pdf-page-size` is `US Letter`. Most of the world uses the `A` format (most commonly `A4`). I would like to discuss, if we could introduce this as the default format in `asciidoctor-pdf`.

Of course, one format has to be the default, and everyone prefers the one that she uses on a daily bases. I think going for the ISO standard is a reasonable choice.

Also, the property that the `A` format scales nicely to bigger and smaller sheets (two `A5` sheets being exactly one `A4`, etc, due to the aspect ratio) makes it superior format (in my completely biased point of view, at least).

Any thoughts on that?

stephenhay:
Resolves #319.

Visually confirmed in Evince: File > Properties.

### DIFF
--- a/lib/asciidoctor-pdf/converter.rb
+++ b/lib/asciidoctor-pdf/converter.rb
@@ -268,7 +268,7 @@ class Converter < ::Prawn::Document
       end
     end
 
-    pdf_opts[:page_size] = (page_size || 'LETTER')
+    pdf_opts[:page_size] = (page_size || 'A4')
 
     pdf_opts[:text_formatter] ||= FormattedText::Formatter.new theme: theme
     pdf_opts
--- a/data/themes/base-theme.yml
+++ b/data/themes/base-theme.yml
@@ -5,7 +5,7 @@ page_background_color: 'FFFFFF'
 page_layout: portrait
 # 36 is equivalent to 0.5in
 page_margin: 36
-page_size: LETTER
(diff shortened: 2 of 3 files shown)


lib/asciidoctor-pdf/converter.rb
```diff
@@ -1,4 +1,4 @@
-# encoding: UTF-8
+# encoding: utf-8
```
veeg: This probably shouldn't be part of this PR.

mojavelinux: I agree. Ruby stores encoding names internally in uppercase, so technically UTF-8 is more correct (though both cases work).

mojavelinux:
Thanks! The change to the page_size is correct. Could you revert the encoding line, amend the commit (`git commit --amend lib/asciidoctor-pdf/converter.rb`), then force push the branch (`git push --force`)?

lib/asciidoctor-pdf/converter.rb
stephenhay: Apologies. My editor changed that automatically and I missed it.

stephenhay:
This has been done.

mojavelinux:
Nice work again @stephenhay!

Issue #342, opened 2015-10-27
Admonition Background Colours/Borders


abradshaw:
Id like to request the ability to set the background colours and/or border settings of each admonition block please

mojavelinux:
#444 is a duplicate of this issue. Since it contains more context, I'm going to close this issue as a duplicate and let that one stand.

PR #379, opened 2015-12-21
Remove hardcoded bold font for first line in abstract

### LINKED ISSUE #378: Harcoded bold for first line of abstract
This means that font styling can't really be used for `abstract`, it will look silly for any setting other than `bold`. For styling of the first line altering the indentation might make sense in some cases, but why a different font style would be used makes no sense to me!

It would be an improvement to simply drop `first_line_options` in this context. It could then be re-introduced with more flexibility and controlled from the theme later on.

Code is here:
https://github.com/asciidoctor/asciidoctor-pdf/blob/master/lib/asciidoctor-pdf/converter.rb#L339

nawroth:
The feature could be re-introduced with theming support later on.
As it stands before this change, it limits the styling of abstract blocks too much.

Fixes #378.

### DIFF
--- a/lib/asciidoctor-pdf/converter.rb
+++ b/lib/asciidoctor-pdf/converter.rb
@@ -336,8 +336,14 @@ class Converter < ::Prawn::Document
     add_dest_for_block node if node.id
     pad_box @theme.abstract_padding do
       theme_font :abstract do
-        # FIXME control first_line_options using theme
-        prose_opts = { line_height: @theme.abstract_line_height, first_line_options: { styles: [font_style, :bold] } }
+        prose_opts = { line_height: @theme.abstract_line_height }
+        # FIXME control more first_line_options using theme
+        if (first_line_font_style = @theme.abstract_first_line_font_style)
+          first_line_font_style = first_line_font_style.to_sym
+          if ( first_line_font_style != font_style)
+            prose_opts[:first_line_options] = { styles: [font_style, first_line_font_style] }
+          end
+        end
         # FIXME make this cleaner!!
         if node.blocks?
           node.blocks.each do |child|
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@
 /pkg/
 /.ruby-gemset
 /.ruby-version
+/.idea
+
--- a/data/themes/default-theme.yml
+++ b/data/themes/default-theme.yml
@@ -125,6 +125,7 @@ abstract:
   font_size: $lead_font_size
   line_height: $lead_line_height
   font_style: italic
+  first_line_font_style: bold
 admonition:
   border_color: $base_border_color
   border_width: $base_border_width


mojavelinux:
There are users that depend on this feature, so we can't remove it outright. The theme setting will need to be part of this change. I would prefer for the default theme to define the bold style for the first line as that is part of the design of the default theme, but I 100% agree it should be possible to control it.

nawroth:
@mojavelinux I wasn't sure about the naming but went with `abstract_firstline_font_style`.
It didn't make sense for me to have `first` in the hierarchy.
Unless we want to have `first_line` and `first_character` in the future ... or something!

mojavelinux:
That's close to the name I had in mind. How about `first_line` instead of `firstline` to be more consistent with CSS.

```
abstract:
  first_line_font_style: bold
```

We don't need it in the base theme, just the default theme. Since it could be absent, can you write it as the following?

```ruby
prose_opts = { line_height: @theme.abstract_line_height }
if (first_line_font_style = @theme.abstract_first_line_font_style) && first_line_font_style != font_style
  prose_opts[:first_line_options] = { styles: [font_style, first_line_font_style] }
end
```

> It didn't make sense for me to have first in the hierarchy.

It doesn't matter. That's just a coding styling. All the levels are flattened using an underscore. I tend to keep it flat until there are at least two keys.

To summarize:

* remove from base theme
* rename to `first_line`
* only pass `:first_line_options` if key is set in theme

Then we'll be good to merge!

mojavelinux:
> Unless we want to have first_line and first_character in the future ... or something!

That's definitely the long-term plan. As much of core CSS that we can replicate, the easier it is on the designer. But we'll take it a chunk at a time.

nawroth:
@mojavelinux Pushed. Your code snippet is missing a `to_sym` !

More than extra options for the first line, I'd like to see borders supported for `abstract` blocks. Maybe it's just a bug around that?

lib/asciidoctor-pdf/converter.rb
```diff
@@ -336,8 +336,11 @@ def convert_abstract node
     add_dest_for_block node if node.id
     pad_box @theme.abstract_padding do
       theme_font :abstract do
-        # FIXME control first_line_options using theme
-        prose_opts = { line_height: @theme.abstract_line_height, first_line_options: { styles: [font_style, :bold] } }
+        prose_opts = { line_height: @theme.abstract_line_height }
+        # FIXME control more first_line_options using theme
+        if (first_line_font_style = @theme.abstract_first_line_font_style.to_sym) && first_line_font_style != font_style
```
mojavelinux: Using `.to_sym` in the assignment will throw an exception if this key is not present. Therefore, the `.to_sym` need to be moved to where `first_line_font_style` is referenced once the truthy check passes.

mojavelinux:
We have support for putting borders around blocks but, like with all things in Prawn, we have to explicitly add the logic. Borders are a bit tricky though because they can span pages, so the abstraction is still a bit weak atm. Similar to #342, but no dedicated issue exists.

lib/asciidoctor-pdf/converter.rb
nawroth: Not having any idea how expensive `to_map` is, I'll go with a nested if.

mojavelinux: Thanks!

The general rule of thumb with Ruby is to avoid method calls if there's another way to do it without one. So the nested if is the right choice in this case (or just call `.to_sym` both times)!

mojavelinux:
Thanks!

PR #406, opened 2016-03-08
Add hints for how to apply styles using Maven and Gradle.

fwilhe:
Note that the Gradle Link does not work until https://github.com/asciidoctor/asciidoctor-gradle-examples/pull/17 is merged.


mojavelinux:
Thanks!

PR #415, opened 2016-03-27
resolves #383 add support for recto/verso margins

### LINKED ISSUE #383: Margin for inner binding (recto/verso margins)
Looking around through the theming website and bug tracker, I do not see a way to set an inner binding offset. I see a margin property that states it is the same as CSS, (eg, left, right, top and bottom.) I do see that the default book theme, `asciidoctor-pdf -d book`, alternates page numbers like a book, but the margins are still the same.

This is useful for books, or even reports, where some of the inner margin is lost due to the binding process (such as those report covers with the plastic rail, you easily lose a quarter inch or more.)

mojavelinux:
- add support for recto/verso margins when attribute media=prepress
- introduce page_margin_inner and page_margin_outer keys to theme
- calculate page margin for recto and verso pages automatically
- add recto/verso support to running content
- start title page, toc, and content on recto page when media=prepress and doctype=book
- add set_page_margin helper
- add page_margin helper
- add page_side helper
- add recto_page? and verso_page? helpers
- document "prepress" publishing mode in theming guide

### DIFF
--- a/data/themes/default-theme.yml
+++ b/data/themes/default-theme.yml
@@ -25,6 +25,9 @@ page:
   background_color: ffffff
   layout: portrait
   margin: [0.5in, 0.67in, 0.67in, 0.67in]
+  # margin_inner and margin_outer keys are used for recto/verso print margins when media=press
+  margin_inner: 0.75in
+  margin_outer: 0.59in
   size: A4
 base:
   align: justify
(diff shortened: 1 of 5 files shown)


mojavelinux:
As currently implemented, the running/header footer does not honor different top and bottom margins between recto and verso pages.

PR #439, opened 2016-05-09
Fix "header" content on verso side

gregturn:
Pasting the "header" YML content doesn't work on the verso side. This patch to the theming guide gives the proper information.


mojavelinux:
Thanks @gregturn!

PR #443, opened 2016-05-17
Uses arrow for menus in pdf output

vogella:
This is basically a guess now to fix that. I don't know how to test this locally.


mojavelinux:
The safest character to use is [\u203a](http://www.fileformat.info/info/unicode/char/203a/index.htm) since that's in the winansi character set. I think it also looks the nicest. I added a little bit of flare to it in the default theme so it stands out. I think you're really going to like it.

I also introduce a key in the theme to control the content used for the caret, so you can set it to whatever character your font supports...in whatever color you want (using pseudo HTML tags):

```
menu_caret_content: " \u25ba "
```

Merged with modifications as 1b530ae40abf2e92412be901ec1199cf7134c0e2. Thanks!

PR #350, opened 2015-11-02
Add support for quoted text attributes (prefixed in square brackets)

Hextremist:
Supports all the types listed in the examples for asciidoc:
http://www.methods.co.nz/asciidoc/chunked/ch10.html
I.e. work for underline, overline, line-through, colors, big and
small.

PR #452, opened 2016-05-24
resolves #451 which suggest to use a stylable background for btn:

### LINKED ISSUE #451: btn: macro should add a background to the button
It would be nice if the PDF output would add a themeable background color to a button. I think the relevant macro is convert_inline_button node. 

I put up a proof of concept patch very soon but unforunately can't find an example in the existing inline macros for that. The other macros seem to follow another approach for styling so advice or update of my little proof of concept patch would be helpful.

Once this is fixed for buttons, I think a fix for improved kbd: support with a background color and a border should be trivial.

vogella:
Currently just a test commit to trigger feedback how to set the background color a text element

### DIFF
--- a/docs/theming-guide.adoc
+++ b/docs/theming-guide.adoc
@@ -1180,7 +1180,8 @@ TIP: The title page can be disabled from the document by setting the `notitle` a
   align: right
 
 |background_color^[1]^
-|<<colors,Color>>
+|<<colors,Color>> +
+(default: _inherit_)
 |title_page:
   background_color: #eaeaea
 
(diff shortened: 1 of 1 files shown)


mojavelinux:
Actually, there's an open pull request for adding a background color to a span of text. The problem isn't that we can't specify the markup, but that the drawing element isn't implemented. See #350. We'll still need this PR, but it is blocked by #350.

mojavelinux:
I think this PR is now superseded by #664. That PR still needs some refinement, but it has all the elements necessary to make this possible.

PR #499, opened 2016-08-07
resolves #481, #301 encode anchors containing characters outside ASCII range in hex

### LINKED ISSUE #481: Items in toc are not clickable if those items contain some CJK characters.
I render this test document with the command ` asciidoctor-pdf -r asciidoctor-pdf-cjk-kai_gen_gothic -a pdf-style=KaiGenGothicCN test.asc` with the help of `asciidoctor-pdf-cjk` gem package which provides a library for a good theme and font solution.

Version info:
```
Asciidoctor PDF 1.5.0.alpha.11 using Asciidoctor 1.5.4 [http://asciidoctor.org]
Runtime Environment (ruby 2.2.5p319 (2016-04-26 revision 54774) [x86_64-cygwin]) (lc:UTF-8 fs:GBK in:- ex:UTF-8)
```

test.asc
```
= test
:doctype:   article
:docinfo:
:toc:
:toclevels: 3
:toc-title: 目录(Table of Contents)


== English clickable

== 中文 Chinese unclickable

== 한국어 Korean clickable

== カタカナ 片仮名 Japanese unclickable
```

You can also get my rendered PDF [here](https://github.com/asciidoctor/asciidoctor-pdf/files/369249/test.pdf).

mojavelinux:
- hex encode anchors that contain characters outside the ASCII range
- squelch the following warning: regexp match /.../n against to UTF-8 string (pdf-core < 0.6.1)

### DIFF
--- a/lib/asciidoctor-pdf/converter.rb
+++ b/lib/asciidoctor-pdf/converter.rb
@@ -313,9 +313,15 @@ class Converter < ::Prawn::Document
       # QUESTION should we store page_start & destination in internal map?
       # TODO ideally, this attribute should be pdf-page-start
       sect.set_attr 'page_start', page_number
-      # NOTE auto-generate an anchor if one doesn't exist so TOC works
       # QUESTION should we just assign the section this generated id?
-      sect.set_attr 'anchor', (sect_anchor = sect.id || %(__autoid-#{page_number}-#{y.ceil}))
+      if (sect_anchor = sect.id)
+        # NOTE encode anchor in hex if it contains characters outside the ASCII range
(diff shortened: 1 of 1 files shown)


mojavelinux:
Technically, the cross references work when using pdf-core 0.6.1. However, since the file stream is written in binary encoding (ASCII-8BIT), it's best that all raw strings be encoded into an ASCII compatible string. Otherwise, the characters are very cryptic when looking at the raw PDF source and thus hard to analyze and test.

mojavelinux:
Use the following to unencode references in the PDF source that begin with 0x.

```
[value[2..-1]].pack 'H*'
```

PR #547, opened 2016-09-10
Fix specification of indentation.

meisterluk:
The given description is faulty.

Let the parent's indentation be 0. Twice the parent's indentation is 0, but "2" is the expected indentation.

To quote the default theme:

```
font:
  catalog:
    NotoSerif:
      normal: notoserif-regular-latin.ttf
```

Consider the indentation of "normal". Its parent indentation is 4. According to the description, indentation 4*2 = 8 would be required, but actually 6 is expected.

### DIFF
--- a/docs/theming-guide.adoc
+++ b/docs/theming-guide.adoc
@@ -147,7 +147,7 @@ base:
     size: 12
 ----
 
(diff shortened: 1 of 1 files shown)


mojavelinux:
Thank you for offering this correction!

Issue #85, opened 2015-01-12
Footnotes are not positioned correctly


gAmUssA:
I trying to use footnotes with pdf like

![image](https://cloud.githubusercontent.com/assets/433085/5711778/1e55f8c4-9a7b-11e4-87d6-64914ab3b807.png)

in pdf it doesn't display correctly

![image](https://cloud.githubusercontent.com/assets/433085/5711795/329858cc-9a7b-11e4-941a-0a93ffe350b5.png)

Thanks

gAmUssA:
Not sure but maybe it relates to #83

mojavelinux:
You should be able to use cross references inside a footnote (keep in mind that footnotes display inline currently). Thus, you can do something like:

```
SPI footnote:[More information about the SPI can be found in <<chapter_13#,Chapter 13>>
```

I recommend using an explicit title in the xref as I have done, though the implicit title should work when creating the PDF.

jmini:
In my opinion this has nothing to do with references but with the fact that footnotes display inline currently:

Example:

    Just the Java code.footnote:[
    With the exception of the [filename]+plugin.xml+ and [filename]+MANIFEST.MF+ files required for Eclipse plugins.
    ]. 
    Developers preferring to write the necessary code manually, may do so. 

PDF:

![pdf output](https://cloud.githubusercontent.com/assets/1222165/7492234/0ca249b8-f3f2-11e4-9d65-1a5d7756a3a7.png)

HTML:

![html output](https://cloud.githubusercontent.com/assets/1222165/7492243/220458f0-f3f2-11e4-91b1-ad1aa2d631a2.png)

Footnotes looks much better in the HTML output because they are not inlined.

mojavelinux:
We inlined the footnotes as a temporary, short term solution. The long term
solution is to put them at the bottom of page or chapter.

The challenge is that we have to reserve space for the footnotes, but we
have to know how much space we need and on which page. Thus, we have to
hook into the page transition, which turns out to be a lot of work. This is
where PDF gets really hard :/

mojavelinux:
If we use the end of chapter, things get a lot simpler (but still some
work).

mojavelinux:
[...]

Issue #73, opened 2014-12-12
Display footnotes at bottom of page instead of inline


mojavelinux:
Display footnotes at the bottom of the page instead of inline.

mojavelinux:
I think for 1.5.0 we can start by displaying them at the end of the chapter. Putting them at the bottom of the page is very difficult task.

rlopez133:
Excited to see this in the beta milestones with the next tag! :+1:

Gitward:
Displaying footnotes seems to work when creating the PDF using https://github.com/asciidocfx/AsciidocFX . What have they done to solve the problem?

My Test-document:

```
= Test
:doctype: article
:notitle:
:!toc:

AsciidocFX shows links in PDFs as footnotes http://stackoverflow.com[SO].

.Asciidoc in PDF does not work in Asciidoctor, but works in AsciidocFX.
[cols="2,5a"]
|===
|Line with Asciidoc code
|here comes a list:

* item 1
* item 2
* item 3

http://stackoverflow.com[Get Answers]!

|Line
|with a footnotefootnote:[footnotes do work in AsciidocFX's PDF output (but not in the preview).]

|===
```
The output of asciidoctor (using gradle):

![asciidoctor](https://cloud.githubusercontent.com/assets/8510849/15852939/532ac8de-2ca3-11e6-8908-821c0bba0895.png)

The output of asciidocFX:

![asciidocfx](https://cloud.githubusercontent.com/assets/8510849/15852949/5ac9f6be-2ca3-11e6-943e-47ba275c07bd.png)

mojavelinux:
AsciiDocFX is not yet using Asciidoctor PDF. Instead, it uses the DocBook toolchain. It's essentially using [fopub](https://github.com/asciidoctor/asciidoctor-fopub) under the covers, which executes and manages the DocBook toolchain.

PierreR:
I was so disappointed when I realized this was not yet implemented ... And would be so happy if it comes up soonish ;-)

mojavelinux:
I'm essentially proposing that we start with treating footnotes as endnotes. I think that's a good first step.

mojavelinux:
For the book doctype, the footnotes should be placed at the end of the chapter. For the article doctype, the footnotes should be placed at the end of the document.

[...]

PR #553, opened 2016-09-11
Colorize footnotes

meisterluk:
**Before**

Footnotes are always rendered completely gray (#999999).

**After**

3 keys are added to the base theme, namely:

```
footnote_reference_font_color: '999999'
footnote_marker_font_color: '999999'
footnote_text_font_color: '999999'
```

which can be overridden by the user in a YAML theme file:

```
footnote:
  reference_font_color: #00FF00
  marker_font_color: #FF0000
  text_font_color: #0000FF
```

Theme extensions might require further considerations such as exchanging square bracket [] with [Japanese 【】](https://www.tofugu.com/japanese/japanese-punctuation/) or alike. But this patch implements a current usecase I have for German text. So this pull request is restricted to _colorizing_ footnotes.

Feel free to discuss 😊

### DIFF
--- a/data/themes/base-theme.yml
+++ b/data/themes/base-theme.yml
@@ -45,6 +45,9 @@ description_list_term_font_style: normal
 description_list_term_spacing: 4
 block_margin_top: 0
 block_margin_bottom: 12
+footnote_reference_font_color: '999999'
+footnote_marker_font_color: '999999'
+footnote_text_font_color: '999999'
 caption_align: left
 caption_font_style: italic
 caption_margin_inside: 4
--- a/docs/theme-schema.json
+++ b/docs/theme-schema.json
@@ -109,6 +109,18 @@
     },
     "base_font_size_small": {
       "type": "number"
+    },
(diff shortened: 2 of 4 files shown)


mojavelinux:
I'm certain we'll need keys in the theme to control the appearance of footnotes, but right now the placement of the footnote text is temporary (just a placeholder). I'd rather spend the time working this into a proper solution for footnotes as described #73 (emitting them as endnotes at the end of the chapter for the book doctype and the end of the document for the article doctype). I'll continue the discussion there.

mojavelinux:
I believe this will be resolved by the fix for #85. If the footnotes are displayed at the end of the chapter, and can be individually styled, do you still think it's necessary to be able to control the style of the footnote marker?

mojavelinux:
Now that #85 has been resolved, I'm going to close this issue. If you'd like to still be able to customize settings which are currently hardcoded, feel free to submit another PR and I'll see that it gets merged.

PR #562, opened 2016-09-18
resolves #386 implement converter for index

### LINKED ISSUE #386: Generate an index
If the document has a section with the style "index", add an index to the output document at that location.

For more details about index creation, see https://github.com/asciidoctor/asciidoctor/issues/450

mojavelinux:
resolves #386 implement converter for index

- implement converter for index
- layout index terms in two columns (with workarounds for cursor handling)
- incrementally build index in convert_inline_indexterm handler
- add anchor points at index terms to reference from index
- render index terms as content for section marked with index style (e.g., [index])
- add pdf-pagenum-offset attribute to document to track number of front matter pages
- use low-level cursor calculation in margin to workaround bug in column_box context
- rename argument to margin methods from y to n

### DIFF
--- a/lib/asciidoctor-pdf/converter.rb
+++ b/lib/asciidoctor-pdf/converter.rb
@@ -20,6 +20,7 @@ require_relative 'pdfmark'
 require_relative 'asciidoctor_ext'
 require_relative 'theme_loader'
 require_relative 'roman_numeral'
+require_relative 'index_catalog'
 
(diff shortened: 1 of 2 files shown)


vogella:
Will this be able to create an index in the PDF?

mojavelinux:
Yes, that is exactly what this does. Here's a preview:

index-sample

vogella:
Awesome.

PR #605, opened 2016-10-02
merge PR #605

### LINKED ISSUE #262: Gem file structure
According to http://guides.rubygems.org/name-your-gem/

asciidoctor-pdf require statement should be

``` ruby
require 'asciidoctor/pdf'
```

Then the file structure should be

```
|-- lib
|   `-- asciidotor
|       |-- pdf
|       |   `-- version.rb
|       `-- pdf.rb
```

And I think module name should be `Asciidoctor::PDF` (current is `Asciidoctor::Pdf`)

I'm going to create some CJK extension(patch and theme) for asciidoctor-pdf, I'm confuse which rule should I follow.

mojavelinux:
To align with RubyGems conventions:
- add asciidoctor/pdf as require alias
- rename Pdf module to PDF
- alias Pdf to PDF for backwards compatibility / convenience
- use uppercase for acronyms in constants

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -8,6 +8,7 @@ For a detailed view of what has changed, refer to the {uri-repo}/commits/master[
 == Unreleased
 
 * drop support for Ruby < 2.3 (and installation will fail for Ruby < 2.1)
+* add asciidoctor/pdf and asciidoctor/pdf/version require aliases
 * switch to tilde dependency versions (flexible patch number) instead of ranges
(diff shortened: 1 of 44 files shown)


mojavelinux:
For what it's worth, Crystal (in many ways a Ruby reboot) stuck with the acronym as uppercase convention. See https://crystal-lang.org/api/0.22.0/

mojavelinux:
This PR handles everything proposed in #262 that's public facing. The one thing that isn't done is to use asciidoctor/pdf as the folder name for all the internal files. However, I didn't want to break the open PRs, so I've decide to defer that change until later (which can be made without breaking anything).

PR #659, opened 2016-11-16
resolves #653 coerce non-nil value of content key from theme to String

### LINKED ISSUE #653: Asciidoctor PDF theme inline content error
I have a theme with the following footer:
```
footer:
  recto_content:
    left: '1/2'
```
It seems the '1/2' is being evaluated as a mathematical expression instead of being rendered as plain text as desired.

`Failed to execute goal org.asciidoctor:asciidoctor-maven-plugin:1.5.3:process-asciidoc (generate-pdf) on project docs: Execution generate-pdf of goal org.asciidoctor:asciidoctor-maven-plugin:1.5.3:process-asciidoc failed: org.jruby.exceptions.RaiseException: (NoMethodError) undefined method ``include?' for 0.5:Float -> [Help 1]`

I tried escaping with a backslash `'1\/2'` but then both slashes are rendered as text.

mojavelinux:


mojavelinux:
I'd like to do this is in theme file instead. After expanding vars, the result should be coersed to a string again. We may need to check for nil.

mojavelinux:
Done.

PR #664, opened 2016-11-20
resolves #451 allow button to be styled, including background and border

### LINKED ISSUE #451: btn: macro should add a background to the button

mojavelinux:
- add support for background-color, border-color, border-width, and border-radius styles for formatted text
- allow button style to be configured by theme
- move default content for button into theme


tinsjourney:
Hi,
any update on this PR ?

mojavelinux:
This PR has been replaced by #1143 (using a simpler approach).

Issue #388, opened 2016-01-09
Duplicate inline image rendering


karanmalhi:
For large images, the image gets rendered twice. 
The first image renders on its own page and is fully rendered. 
The second image shows up on the next page, left aligned and is only partially visible i.e. the left half is missing. 
Using Prawn

mojavelinux:
Can you please provide a link to an image that can be used for testing? The original image size is ideal since we don't want the problem to be lost in transfer.

ShaKeSPeaR:
A little example for issue (see included zip):
 * big_image_test.txt - example project 
 * big_image.png - original image [1274x1215]
 * image.png - a little scale of original image  [944x900]
 * small_image.png - half scale of original image [500x477]

As you may see in big_image_test.pdf:
* big image not show at all
* middle image have twice rendered (with strange crop)
* small image render as expexted

[big_image_zip.zip](https://github.com/asciidoctor/asciidoctor-pdf/files/94331/big_image_zip.zip)

Some remarks:
 * If I manually set a width to fit the image into the page (`image:big_image.png[width="100"]`) it scaled well and render normally
 *  I played with image sizes - in some cases it may render thrice.

roentgenium:
I also encounter the same issue.
And the problem may be using inline image "image:" (one colon) instead of "image::" ([two colon](two colon)).

mojavelinux:
As @roentgenium points out, you are using an inline image when you mean to use a block image. Use two colons after image in the image macro instead of one. (e.g., `image::filename.png[]`).

Inline images are meant to fit without the line of text (i.e., the line height). If they exceed that area, then the behavior is undefined.

mojavelinux:
I figured out why this is happening. It has to do with how we arrange inline images. We first put blank spaces where the image will go. Then we come back and lay the image into the place. However, if that blank space wraps, then we end up displaying the image multiple times, once for each line. (We also have a rounding error that causes this blank space to be too large in certain conditions).

We need to figure out some way to detect the end of the line and only allow the image to be as wide as the remaining space on the line (or wrap it to the next line so it does have enough space).

We may have to solve this one in stages since there are several problems at play here.

mojavelinux:
This is related to #620.

mojavelinux:
Actually, this is simpler than I thought. The spacer is wider than the available width and Prawn attempts to render the image again on each wrapped line. We just need to ensure the spacer is no wider than a single line.

[...]

PR #672, opened 2016-11-25
resolves #620 Add support for pdfwidth to inline images

### LINKED ISSUE #620: Add support for pdfwidth to inline images
pdfwidth does not work for an inline image.

```
image:model_spy_targetplatform.png[pdfwidth=1%]
```

image

I upload soon an example via a PR.

fap-:


mojavelinux:
This is great, except I don't think we should support vw units or the theme fallback. vw units don't make sense in this context (and require special handling anyway). The theme fallback is intended for block images only. Inline images should be sized explicitly.

Here's how I propose to write the line.

```ruby
width_attr = (width = resolve_explicit_width node.attributes, bounds.width) ? %( width="#{width}") : nil
```

mojavelinux:
I've rewritten the PR and pushed it to the PR branch. Take a look and see what you think.

Here's a summary of what I changed:

* I create a `preresolve_explicit_width` image to more clearly describe that we are doing a partial resolution of the explicit width by resolving absolute units, but not relative ones
  - I also included support for the `scaledwidth` attribute
* I simplified the resolution of percentage value in the arranger
* I constrained the width of the image to the width of the bounds (in this context, a image larger than the available width doesn't make sense and causes Prawn to loop infinitely)

If you set the pdfwidth to 100%, we still run into the issue described in #388. However, I now understand why that is happening and can fix it. The spacer is wider than the available width and Prawn attempts to render the image again on each wrapped line.

fap-:
Wow, thank you, looks good to me.

My hack to fix #388 would look like this, but there is probably a better solution. (with the prawn api that you are mentioning in the comment?)

```
        if ((number = (fragment[:image_width] / spacer_w).ceil) * spacer_w) < doc.bounds.width
          fragment[:text] = ImagePlaceholderChar * number
        else
          fragment[:text] = ImagePlaceholderChar * (fragment[:image_width] / spacer_w).floor
        end
```

mojavelinux:
Indeed! That's almost identical to what I was going to propose, so we're on the same wavelength.

```ruby
if (spacer_cnt = (fragment[:image_width] / spacer_w).ceil) * spacer_w > available_w
  spacer_cnt -= 1
end
fragment[:text] = ImagePlaceholderChar * spacer_cnt
```

We know if it's too much, it will be exactly 1 too many.

There's really no other way I can think of to fix it unless we complete rethink how space is allocated. And for that to happen would probably require a change to Prawn or to patch a significant part of it. We might be able to get some more fine tuning using character spacing, but I'll need to experiment to know if that helps us.

PR #682, opened 2016-12-07
resolves #388 Duplicate inline image rendering

### LINKED ISSUE #388: Duplicate inline image rendering

fap-:
- decrement amount of placeholder characters if their width exceeds max width

I took your proposal because it felt 'more correct' and intuitive.
I removed the inline assignment since I felt it makes the code harder to read than it has to be.


mojavelinux:
This looks great. Thanks!

PR #687, opened 2016-12-08
resolves #686 use character spacing to fine-tune width of placeholder for inline image

### LINKED ISSUE #686: Large inline image spills outside the content area to the left
A large inline image (for example, an image that has pdfwidth=100%) spills outside the content area to the left. This can be seen by stacking a block image above an inline image, both having the same dimensions.

```
image::flag.svg[pdfwidth=100%]

image:flag.svg[pdfwidth=100%]
```

The inline image appears to the left of the block image.

This happens because of how space is reserved for inline images. We repeat a placeholder character across the line to reserve space for the image. However, that only provides a certain precision. If the width of the placeholder is shorter than the width of the image, the image appears shifted to the left because it is centered on the shorter span.

We need to ensure that the placeholder width is exactly the same as the image width. This can be done by applying a character width to the placeholder text that distributes the extra space across each character.

If the width of the image is not wide enough for a single placeholder character, we should force it to use a single placeholder and then set a negative character spacing to recover the space.

mojavelinux:
- use character spacing to distribute extra space not covered by placeholder chars
- force at least one placholder char for narrow images and use negative character spacing to recover overrun
- simplify how spacer count is computed

### DIFF
--- a/lib/asciidoctor-pdf/formatted_text/inline_image_arranger.rb
+++ b/lib/asciidoctor-pdf/formatted_text/inline_image_arranger.rb
@@ -97,7 +97,7 @@ module InlineImageArranger
         spacer_w = nil
         doc.fragment_font fragment do
           # NOTE if image height exceeds line height by more than 1.5x, increase the line height
-          # HACK we could really use a nicer API from Prawn here; this is an ugly hack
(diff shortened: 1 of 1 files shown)


mojavelinux:
Depends on #685.

Issue #445, opened 2016-05-17
AsciiDoc table cell styling not functioning in PDF output


royemmerich:
# Version

```
$ asciidoctor-pdf -V
Asciidoctor PDF 1.5.0.alpha.11 using Asciidoctor 1.5.4 [http://asciidoctor.org]
Runtime Environment (ruby 2.0.0p648 (2015-12-16 revision 53162) [universal.x86_64-darwin15]) (lc:UTF-8 fs:UTF-8 in:- ex:UTF-8)
```
# My Wish

To insert an unordered list in the cell of a table in a PDF file.
# The Problem

The table below is a direct copy from the code block example entitled:

**Comparing cells using AsciiDoc styles and no AsciiDoc styles**

from the [asciidoc table cell formatting section](http://asciidoctor.org/docs/user-manual/#cell) of the Asciidoctor Manual.

``` asciidoc
[cols="2"]
|===

a|This cell is prefixed with an `a` so the following list is rendered with the AsciiDoc list element styles.

* List item 1
* List item 2
* List item 3
|This cell *is not* prefixed with an `a` so the following list is not rendered with the AsciiDoc list element styles.

* List item 1
* List item 2
* List item 3

a|This cell is prefixed with an `a` so the following paragraph is rendered with the `lead` style.

[.lead]
I am a paragraph styled with the lead attribute.
|This cell *is not* prefixed with an `a` so the following paragraph is not rendered with the `lead` style.

[.lead]
I am a paragraph styled with the lead attribute.

|===
```

When I generate the PDF the end result looks like this:
![screen shot 2016-05-17 at 11 30 12](https://cloud.githubusercontent.com/assets/3904395/15317673/c84e721e-1c22-11e6-8c27-ab8660db1e97.png)

when it should look like this (screenshot from Asciidoctor manual):

![screen shot 2016-05-17 at 11 32 16](https://cloud.githubusercontent.com/assets/3904395/15317738/0e3d05e2-1c23-11e6-9d75-c25c24fb59ea.png)

mojavelinux:
This is a known limitation right now. See #6.

mojavelinux:
As I mention in #6, what we're going to need to do is implement a custom cell handler in Prawn. That's the only way we'll be able to delegate back into the converter so that the page calculations can be handled by Prawn.

Issue #358, opened 2015-11-17
AsciiDoc Content in Tables Do Not Render (Unordered Lists)


jaredmorgs:
I've noticed that unordered lists in tables don't seem to render correctly.

If I have the following structure:

```
[[personal-mobile]]
.User Expectations
[cols="2a,2a"]
|===
|Expected or Permitted Behaviors |Forbidden Behaviors

|
Immediately inform your line manager if:

* your device is lost.
* your device is stolen.
* you believe any of your accounts (business and personal) you access from that device have been compromised.

|
Wait for a few days to see if someone hands in your personal device before you let your line manager know it is missing.

|===
```

The list renders like this in the PDF:

<img width="827" alt="screenshot 2015-11-17 11 30 45" src="https://cloud.githubusercontent.com/assets/952793/11200102/b5e9cf08-8d1e-11e5-9e44-272bd1a5ef7d.png">

veeg:
AsciiDoc cell content (including lists), which is enabled in AsciiDoc with the `e` prefix to the cell divider, e.g.,
```
|===
e| This cell supports AsciiDoc content!
|===
```

is not currently supported in Asciidoctor-pdf. 
Se main issue #6, and duplicate issues reported #231 and #315.

mojavelinux:
As @veeg points out, these are known issues. We need to create a custom cell type in Prawn in order to support complex (aka AsciiDoc) content in table cells.

Issue #315, opened 2015-09-09
How do I create lists within a table cell?


rlopez133:
What I currently have:

```
|====
^|Revision ^|Release Date ^|Author(s) ^|Notes

^| 1.1 ^| Wednesday Sept 9, 2015 ^| Joe Talerico, Roger Lopez ^a| * Grammatical error and acronym fixes 
^| 1.0 ^| Friday July 17, 2015 ^| Joe Talerico, Roger Lopez ^a| * Initial Release
4+^.e| PDF generated by https://github.com/asciidoctor/asciidoctor-pdf[Asciidoctor PDF]
|====
```

I read in the syntax guide that all I needed to put was an "a" in front of the cell and that would work but its not working for me. I'm sure I'm just doing it wrong.

mojavelinux:
AsciiDoc table content still isn't supported yet. That includes lists, so it's covered by #6.

Issue #650, opened 2016-11-01
Can I use a locally produced gem in my Gradle build


vogella:
@fap- is currently build the Asciidoctor-pdf gem so that we can benefit from not yet released versions.

@mojavelinux describes how to use a released Ruby version of asciidoctor in https://github.com/asciidoctor/asciidoctor-pdf/issues/572#issuecomment-248848703

Is it also possible to use a locally produced gem in a Gradle build? We tried to use flatdir but failed and our attempts to put it into the local Gradle cache or .m2 cache were also not successfully.

fap-:
Hey there,

so far we tried to get the gradle jruby to load the development version of asciidoctor-pdf from a local maven or gradle repository/cache, but we weren't successful.

Another option would be to locally host the gem with rubygems_servlets, but this feature is [undocumented](https://github.com/torquebox/rubygems-servlets/issues/15) so far.

Would it be alright with you if we uploaded the current asciidoctor-pdf dev version to rubygems.org, under our own 'namespace'? e.g. 'vogella-asciidoctor-pdf-dev'
This would probably solve our problem.

mojavelinux:
I would really prefer if you didn't add a development gem to rubygems.org since it will stick around forever and eventually confuse people.

There is a way to specify a path to a local directory or file as the gem dependency using the JRuby Gradle plugin. If you figure this out, it will help out a lot of other people who have tried. We can then document it in the Gradle plugin README so we don't forget.

mojavelinux:
See https://github.com/jruby-gradle/jruby-gradle-plugin/issues/102

mojavelinux:
I figured out how you can do this. It's a bit of a hack, but it should get the job done for now.

First, you need to build the gem locally and store it somewhere.

```
git clone https://github.com/asciidoctor/asciidoctor-pdf.git
cd asciidoctor-pdf
bundle
bundle exec rake build
```

The gem file can be found under pkg/.

We'll need to manipulate the configuration for the JRuby Gradle plugin, so you need to declare it:

```
plugins {
  ...
  id 'com.github.jruby-gradle.base' version '1.3.3'
}
```

In order to use a local gem from Gradle, we have to specify all the dependencies of that gem manually.

```
dependencies {
  gems 'rubygems:prawn:2.1.0'
  gems 'rubygems:prawn-table:0.2.2'
  gems 'rubygems:prawn-templates:0.0.3'
  gems 'rubygems:prawn-svg:0.25.2'
[...]

PR #707, opened 2016-12-18
resolves #6 add support for AsciiDoc table cells (including nested tables)

### LINKED ISSUE #6: Add support for AsciiDoc cell content
Add support for AsciiDoc cell content. Currently, the raw AsciiDoc source is passed through to the PDF file.

mojavelinux:
- use custom Cell implementation to support AsciiDoc table cells
- supports colspan, rowspan and vertical alignment
- works best if column is assigned an explicit (relative) width
- content that exceeds height of one page is clipped (general limitation of table support in Prawn)
- extra padding from last block is added to bottom of cell
- does not (yet) inherit font properties from table cell

Also resolves duplicates #315, #445 and #358.

### DIFF
--- a/lib/asciidoctor-pdf/converter.rb
+++ b/lib/asciidoctor-pdf/converter.rb
@@ -1460,16 +1460,14 @@ class Converter < ::Prawn::Document
           cell_data[:content] = preserve_indentation cell.text, (node.document.attr 'tabsize')
           cell_data[:inline_format] = true
         when :asciidoc
-          unless defined? @asciidoc_cell_warning
-            @asciidoc_cell_warning = true
-            warn 'asciidoctor: WARNING: The \'a\' (AsciiDoc) table cell style is not supported by this converter. The source of these cells will be treated as paragraph content.'
-          end
-          # TODO finish me
+          asciidoc_cell = ::Prawn::Table::Cell::AsciiDoc.new self,
+              (cell_data.merge content: cell.inner_document, font_style: (val = theme.table_font_style) ? val.to_sym : nil)
+          cell_data = { content: asciidoc_cell }
         else
           cell_data[:font_style] = (val = theme.table_font_style) ? val.to_sym : nil
         end
         unless cell_data.key? :content
-          # NOTE effectively the same as calling cell.content
+          # NOTE effectively the same as calling cell.content (should we use that instead?)
           # TODO hard breaks not quite the same result as separate paragraphs; need custom cell impl
           if (cell_text = cell.text).include? LF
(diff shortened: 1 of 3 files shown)


barspi:
wow... I was having exactly this problem with lists in cells and I've seen it has been resolved 25 minutes ago! :-)
Problem is that the last published version seems to bee alpha.11 and the current head has sooo many fixes and improvements since alpha.11

mojavelinux:
The last published version of the Ruby gem is alpha.13. The last published version of AsciidoctorJ PDF jar is alpha.11. We're working on getting that deployment pipeline automated so that it's kept up to date with the Ruby version. Until then, I recommend configuring your build to use the gem via the JRuby plugin (for example, the JRuby Gradle plugin). See https://github.com/asciidoctor/asciidoctor-pdf/issues/572#issuecomment-248848703 or https://github.com/asciidoctor/asciidoctor-pdf/issues/650#issuecomment-258338060).

barspi:
Yeah, I saw the difference in published versions after I commented (I'm new to asciidoctor)
I could go and build it myself, but I'm developing some public project using a gradle build, so I need an "official" release that gradle can pull down from official repos

I'd like to help but have no idea how :-/

mojavelinux:
A release of alpha.14 is coming very soon. I'm just in the process of wrapping up a few more issues.

Here's the thread about the new AsciidoctorJ PDF repo. http://discuss.asciidoctor.org/New-repo-for-asciidoctorj-pdf-td4988.html Once that gets sorted out, then the releases should remain in sync.

oddhack:
Is there documentation that would explain to a complete Ruby newbie, like me, how I would go about installing the current github in my Ruby environment? I need the fix for this to make progress on converting some of our docs, and the published alpha13 doesn't have it.

barspi:
@oddhack 
I think this may help you  https://github.com/asciidoctor/asciidoctor-pdf/issues/650#issuecomment-258338060

oddhack:
@barspi thanks! The build instructions are helpful but the other stuff about jruby / gradle doesn't seem directly relevant since I'm not using jruby (AFAIK - just the basic ruby-2.3.3 and related command-line tools accessed via rbenv). I guess once I have the 'gem' there's probably some way to bash on the 'gem install' command to access the local resource. This stuff is all pretty mystifying at my current state of understanding, compared to Debian package management. I'm sure it's wonderful once enough Ruby Kool-Aid has been consumed :-)

mojavelinux:
It's my top priority to get alpha.14 released by this weekend. I got help up by a major workload in Jan that I'm just coming off of. Stay tuned...but don't hesitate to nudge me ;)

oddhack:
That will be awesome. Looking forward to stress-testing asciidoc blocks in table cells!

PR #735, opened 2017-01-17
Fixes #259 - Sidebars now render on multiple pages

### LINKED ISSUE #259: Sidebar doesn't render more than 1 page

book pdf_2015-07-06_10-14-09

TobiasHector:
Refactored convert_sidebar to mirror behaviour of convert_quote_or_verse, such that multiple pages draw correctly. Requires drawing the block content twice, and there's a small hack, but it works.

I coded this just to get something working, but happy to accept edits - I am using as-is at the moment and seems to work as expected.

Fixes #259

### DIFF
--- a/lib/asciidoctor-pdf/converter.rb
+++ b/lib/asciidoctor-pdf/converter.rb
@@ -698,18 +698,52 @@ class Converter < ::Prawn::Document
     add_dest_for_block node if node.id
     theme_margin :block, :top
     keep_together do |box_height = nil|
-      if box_height
-        float do
-          bounding_box [0, cursor], width: bounds.width, height: box_height do
-            theme_fill_and_stroke_bounds :sidebar
-          end
-        end
-      end
+      start_page_number = page_number
+      start_cursor = cursor
+      # FIXME? Have to draw graphics once before content to get accurate box_height.
+      # This is overwritten by the theme_fill_and_stroke_bounds command, so needs drawing twice
       pad_box @theme.sidebar_padding do
         if node.title?
           theme_font :sidebar_title do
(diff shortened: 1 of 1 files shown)


TobiasHector:
Closing this - looks like it causes NaNs in the output PDF, which Chrome PDF viewer wasn't catching. A proper fix is going to need some more looking at... once/if I figure out how in the heck to make this work, I'll open a new pull request.

mojavelinux:
Great! I still plan to review your code to see if I can provide some guidance. I apologize for letting this linger. I've been behind on reviews lately.

TobiasHector:
Awesome thanks - no problem with the lingering. FYI if you take out the "HACK" in this code, which only sort of worked anyway, the specific symptom is as follows for a sidebar that spans three pages (the most interesting case):

* First draw of content renders correctly.
* Draw of the box skips the second page, and draws the second and third page boxes on the third page
* Second draw of content also skips over second page, draws second and third page of content on the third page

The whole thing of moving the second draw down an extra page has me scratching my head; I can only assume that there's some lingering state set in the PDF by the first content draw, but I can't figure out what it is. It's not impossible that this is triggered by earlier non-spanning sidebars as well. I'm still trying to work down to a minimal repro, but I'm at a conference this week so won't make progress for a little while.

TobiasHector:
@mojavelinux New fix which doesn't break everything is in #780. Something was getting modified in the graphics state that messed things up for subsequent drawing - saving and restoring graphics state made it work.

TobiasHector:
@mojavelinux Ignore that last one, I got overexcited apparently. #781 *actually* fixes things.

PR #780, opened 2017-03-20
Fixes #259 - Sidebars now render on multiple pages

### LINKED ISSUE #259: Sidebar doesn't render more than 1 page

TobiasHector:
This adds code to render sidebars over multiple pages correctly. It does result in text existing in the document twice (as the last fix did - https://github.com/asciidoctor/asciidoctor-pdf/pull/735), which is a bit annoying, but it only really affects text searches and document size afaict.

A better fix is theoretically possible, but it would require significant reworking of the pdf renderer - the using the "dry_run" attribute, for instance.

Fix refactors convert_sidebar to mirror behaviour of convert_quote_or_verse, such that multiple pages draw correctly. Requires drawing the block content twice.

Fixes #259

### DIFF
--- a/lib/asciidoctor-pdf/converter.rb
+++ b/lib/asciidoctor-pdf/converter.rb
@@ -694,28 +694,68 @@ class Converter < ::Prawn::Document
   alias :convert_quote :convert_quote_or_verse
   alias :convert_verse :convert_quote_or_verse
 
+  def convert_contents_sidebar node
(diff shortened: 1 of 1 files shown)


TobiasHector:
Gah, false alarm, accidentally looking at the wrong spec build... getting closer though!

Issue #782, opened 2017-03-22
error: <top (required)>': can't modify frozen Array (RuntimeError)


fly7632785:
I have followed the Get Start.but some problems happened.
I just wanna whether the reason is for the Ruby's version too high  and  incompatible?
environment:
ruby 2.4

error:
`JafirdeMac-mini:~ jafir$ asciidoctor-pdf -v
/usr/local/lib/ruby/gems/2.4.0/gems/prawn-templates-0.0.3/lib/prawn/templates.rb:246:in `<top (required)>': can't modify frozen Array (RuntimeError)
	from /usr/local/Cellar/ruby/2.4.0/lib/ruby/2.4.0/rubygems/core_ext/kernel_require.rb:55:in `require'
	from /usr/local/Cellar/ruby/2.4.0/lib/ruby/2.4.0/rubygems/core_ext/kernel_require.rb:55:in `require'
	from /usr/local/lib/ruby/gems/2.4.0/gems/asciidoctor-pdf-1.5.0.alpha.14/lib/asciidoctor-pdf/converter.rb:10:in `<top (required)>'
	from /usr/local/lib/ruby/gems/2.4.0/gems/asciidoctor-pdf-1.5.0.alpha.14/lib/asciidoctor-pdf.rb:3:in `require_relative'
	from /usr/local/lib/ruby/gems/2.4.0/gems/asciidoctor-pdf-1.5.0.alpha.14/lib/asciidoctor-pdf.rb:3:in `<top (required)>'
	from /usr/local/Cellar/ruby/2.4.0/lib/ruby/2.4.0/rubygems/core_ext/kernel_require.rb:120:in `require'
	from /usr/local/Cellar/ruby/2.4.0/lib/ruby/2.4.0/rubygems/core_ext/kernel_require.rb:120:in `require'
	from /usr/local/lib/ruby/gems/2.4.0/gems/asciidoctor-pdf-1.5.0.alpha.14/bin/asciidoctor-pdf:4:in `<top (required)>'
	from /usr/local/bin/asciidoctor-pdf:22:in `load'
	from /usr/local/bin/asciidoctor-pdf:22:in `<main>'`

dduportal:
Hello @fly7632785 !

This seems to be corrected in the next version (at least it is merged in master).

@mojavelinux have provided explanation and a quick fix here: #772 

Hope it helps !

fly7632785:
Hi,I've try the answer .

[...]

PR #785, opened 2017-03-26
Stick to Prawn@2.1.x

oncletom:
refs #782


mojavelinux:
This has already been resolved. I just need to do the release, which I'm working on now.

oncletom:
Perfect 👍  I was not sure it was being addressed outside of the #782 issue :-)

mojavelinux:
Not a problem. We definitely got bitten by that upgrade, but I've put in safeguards to prevent it from happening again.

PR #792, opened 2017-03-29
resolves #791 fix layout error causes by nested keep together blocks

### LINKED ISSUE #791: A nested block that uses keep together makes layout go haywire
If a keep together block contains another keep together block (such as a listing block inside of a sidebar block), and the nested block is shifted to another page, the layout goes haywire. This is due to the fact that the keep_together method calls `start_new_page` instead of `advance_page`, as @TobiasHector figured out.

mojavelinux:
* use advance_page in keep_together instead of start_new_page


mojavelinux:
cc: @TobiasHector

TobiasHector:
👍

Looks good to me.

PR #838, opened 2017-08-19
resolves #296: pass 'leading' attribute from theme for table cells to prawn

### LINKED ISSUE #296: Set line height of text in table cell
The the line height (specified using the "leading" property) on the text in a monospaced table cell. Right now, it's using a line height of 1, which looks pretty bad when the line wraps.

madmas:
with the help @fliiiix during a great hacking session at @froscon 12 (thanks a lot!), we found the right place to add the handling for this property and fix #296 .

It helps not only with the line spacing but also to handle appearently broken fonts that seems to report their height incorrectly, like shown in the screenshots:

Before:
screenshot_before

After:
screenshot_after

Also added some example files to demonstrate it. 
Unfortunately, I cannot add the font with the wired behaviour as it's a licensed one :(


mojavelinux:
I implemented this in https://github.com/asciidoctor/asciidoctor-pdf/commit/1a0b00cadd9e2f674d514c99698e953f1d215782. It's a little more complicated than just setting the padding, as you can see.

I didn't add the table example since I'd rather use the chronicles example for that. I would like to have sample themes, but they either belong in the (future) test suite or a dedicated repository.

PR #892, opened 2018-03-08
Catch fw attribute of an icon

### LINKED ISSUE #890: Font Awesome fixed width icons not rendering
Example of source
```asciidoc
Icon 1: icon:magic[]
Icon 2: icon:magic[fw]
```

In the above example only the first icon is generated. The second one, with `fw` (fixed width) set, displays no icon at all.

This doesn't look to be an intentional behavior. If fixed width cannot be applied in PDF rendering the flag should be ignored.

berzley:
resolves #890 
Added output of icon sizing to warning message of invalid icons

This may be along the lines of a 'hack' solution as it doesn't provide fixed width. It should however allow the use of `fw` in the document without upsetting the PDF render.

### DIFF
--- a/lib/asciidoctor-pdf/converter.rb
+++ b/lib/asciidoctor-pdf/converter.rb
@@ -1971,6 +1971,7 @@ class Converter < ::Prawn::Document
       icon_set = 'fa' unless IconSets.include? icon_set
       if node.attr? 'size', nil, false
         size = (size = (node.attr 'size')) == 'lg' ? '1.3333em' : (size.sub 'x', 'em')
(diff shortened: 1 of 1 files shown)


mojavelinux:
I have a counter proposal here #895.

The only think I can't figure out yet is how to center the glyph within the reserved width. Prawn doesn't seem to provide a way to adjust the left position of a character in a line of text.

mojavelinux:
Got it!

berzley:
#895 Looks great, a much better solution!

mojavelinux:
It took quite a bit more code, but I think it's worth it.

PR #895, opened 2018-03-09
resolves #890 add support for fw icons

### LINKED ISSUE #890: Font Awesome fixed width icons not rendering

mojavelinux:
- allow width attribute to be set on font tag
- override width method on Fragment to support explicit width
- set width attribute to 1em if size of icon is fw
- align icon in center of 1em space


mojavelinux:
Thanks for getting this started, @berzley!

PR #923, opened 2018-07-12
Add possibility for recto/verso background images

1marc1:
Created the possibility to add...

* page_background_image_recto
* page_background_image_verso

...to the theme .yml file. This allows the user to have alternating background images.

The proposed change is backwards compatible with the current `page_background_image`. The user can either use the original or the newly proposed method. If the same background image is specified for both the recto and verso image, then the result is the same as if the user had only used `page_background_image`: all pages will have the same background image.


mojavelinux:
I want to respond to a few points that came in via a private e-mail.

> The next challenge is to ensure the code is updated to work with the new layout feature. We probably need to expand on the existing PR.

That will drastically expand the scope of this issue, which I don't want to do. If you require that capability, please create a separate issue.

> It currently works for portrait documents only. I would imagine it to work for landscape-only documents, although I haven't tried.

It will work with whatever layout you start with. In fact, it will still work even if the layout is changed. The background just won't fit the same on those pages.

> Now that the ability exists to change page orientation inside a document, we would need to cater for this by providing users the option to specify something like recto-portrait, recto-landscape, verso-portrait and verso-landscape.

I'll be frank, this is just going outside the scope of what Asciidoctor PDF was designed to do. What's next? Changing the fonts and colors for recto / verso with different layouts? The permutations are just going to explode. If those are your requirements, I challenge whether Asciidoctor PDF is even the right solution. We have to draw the line somewhere.

> If you want I can have a look to see if I can expand on what I have done in the PR to make that work. 

Please use a separate PR that builds on the outcome of this one. If you do add this capability, please prefer using a function / method to avoid exploding the size of the existing method(s).

PR #925, opened 2018-07-27
fix table caption position when alignment is center

### LINKED ISSUE #823: Table caption is not be centered when the table is center.
I want a table and the caption to be center, so I write Asciidoc as follows.

```
.Table caption
[.center%header%autowidth]
|=======================
| H1 | H2 | H3
| Item1 | Item2 | Item3
|=======================
```

And I set align of caption to center in theme.yml
```
caption:
  align: center
```

This works fine when I use asciidoctor-pdf v1.5.0 alpha12.

But when I use asciidoctor-pdf v1.5.0 alpha13 or later, the caption is placed right from the center.

I think the fix of #469 is related. 
Is the caption placed in center between left edge of the table and right edge of the page?

eduidl:
related issue #823

style.yml
```yml
caption:
  align: center
```

### before
image

### after 
image


mojavelinux:
Closing in favor of #1139, which will address this problem.

PR #939, opened 2018-10-08
resolves #891 upgrade to Font Awesome 5

### LINKED ISSUE #891: Prawn Icon v2.0.0 Update
[prawn-icon](https://github.com/jessedoyle/prawn-icon) has been updated to support Font Awesome 5 in the latest buid.

Asciidoctor PDF could possibly be upgraded to use prawn-icon v2.0.0 to utilize supported Font Awesome 5 icons.

jessedoyle:
Happy Hacktoberfest! This is a PR that I've wanted to make for some time now.

* Update `prawn-icon` to version `2.3.0`. This update upgrades
  FontAwesome from version 4 to 5.
* Update the list of supported icon fonts to include the
  FontAwesome categories (`solid => fas`, `brands => fab`
  and `regular => far`).
* The `fa-*` prefix may still be used to generate icons,
  but a deprecation warning will be printed prompting to
  use the Font Awesome 5 font categories.
* Setup `fas` to be the default font family (FontAwesome
  solid).
* Remove all references to the `octicon` font. It is no
  longer included in `prawn/icon`.
* Update documentation and guides that are applicable to
  icon styling.
* Regenerate the `examples/chronicles-example.pdf` file to
  include new icons.

This should resolve https://github.com/asciidoctor/asciidoctor-pdf/issues/891.

### DIFF
--- a/lib/asciidoctor-pdf/converter.rb
+++ b/lib/asciidoctor-pdf/converter.rb
@@ -1079,6 +1079,12 @@ class Converter < ::Prawn::Document
     end
 
     if marker
+      if marker_style[:font_family] == 'fa'
+        logger.info { 'deprecated fa icon set found in theme; use fas, far, or fab instead' }
+        marker_style[:font_family] = FontAwesomeIconSets.find do |candidate|
+          (::Prawn::Icon::FontData.load self, candidate).yaml[candidate].value? marker
+        end || 'fas'
+      end
       marker_gap = rendered_width_of_char 'x'
       font marker_style[:font_family], size: marker_style[:font_size] do
         marker_width = rendered_width_of_string marker
@@ -1260,7 +1266,7 @@ class Converter < ::Prawn::Document
     add_dest_for_block node if node.id
     theme_margin :block, :top
     audio_path = node.media_uri(node.attr 'target')
-    play_symbol = (node.document.attr? 'icons', 'font') ? %(<font name="fas">#{make_icon('fas-play').unicode}</font>) : RightPointer
+    play_symbol = (node.document.attr? 'icons', 'font') ? %(<font name="fas">#{(::Prawn::Icon::FontData.load self, 'fas').unicode 'play'}</font>) : RightPointer
     layout_prose %(#{play_symbol}#{NoBreakSpace}<a href="#{audio_path}">#{audio_path}</a> <em>(audio)</em>), normalize: false, margin: 0, single_line: true
     layout_caption node, side: :bottom if node.title?
     theme_margin :block, :bottom
@@ -1296,7 +1302,7 @@ class Converter < ::Prawn::Document
     if poster.nil_or_empty?
       add_dest_for_block node if node.id
       theme_margin :block, :top
-      play_symbol = (node.document.attr? 'icons', 'font') ? %(<font name="fas">#{make_icon('fas-play').unicode}</font>) : RightPointer
+      play_symbol = (node.document.attr? 'icons', 'font') ? %(<font name="fas">#{(::Prawn::Icon::FontData.load self, 'fas').unicode 'play'}</font>) : RightPointer
       layout_prose %(#{play_symbol}#{NoBreakSpace}<a href="#{video_path}">#{video_path}</a> <em>(#{type})</em>), normalize: false, margin: 0, single_line: true
       layout_caption node, side: :bottom if node.title?
       theme_margin :block, :bottom
@@ -2043,9 +2049,9 @@ class Converter < ::Prawn::Document
       if (icon_name = node.target).include? '@'
         icon_name, icon_set = icon_name.split '@', 2
       else
-        icon_set = node.attr 'set', (node.document.attr 'icon-set', 'fas'), false
+        icon_set = node.attr 'set', (node.document.attr 'icon-set', 'fa'), false
       end
-      icon_set = 'fas' unless IconSets.include? icon_set
+      icon_set = 'fa' unless IconSets.include? icon_set
       if node.attr? 'size', nil, false
         case (size = node.attr 'size')
         when 'lg'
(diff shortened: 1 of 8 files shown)


jessedoyle:
Hey @mojavelinux - I'm wondering if you've had a chance to look at this diff? I'd be happy to make any changes necessary to get this merged!

mojavelinux:
@jessedoyle I finally had a chance to look at this patch. I'm ready to make this upgrade. I'm not worried about the fact that octicons has to be dropped since it's really FontAwesome that we care about.

I fixed the trailing whitespace to make this patch cleaner. Can you rebase?

lib/asciidoctor-pdf/converter.rb
```diff
@@ -2043,7 +2041,9 @@ def convert_inline_icon node
       end
       begin
         # TODO support rotate and flip attributes
-        %(<font name="#{icon_set}"#{size_attr}>#{::Prawn::Icon::FontData.load(self, icon_set).unicode icon_name}</font>)
+        key = [icon_set, icon_name].join('-')
+        data = make_icon(key).format_hash
```
mojavelinux: is make_icon coming from Prawn::Icon?

lib/asciidoctor-pdf/prawn_ext/extensions.rb
```diff
@@ -191,7 +191,7 @@ def register_font data
   def font name = nil, options = {}
     if name
       options = { size: options } if ::Numeric === options
-      if IconSets.include? name
+      if name != 'fa' && IconSets.include?(name)
```
mojavelinux: It's not clear to me why "fa" is being excluded here. Could you explain?

lib/asciidoctor-pdf/prawn_ext/extensions.rb
```diff
@@ -862,6 +862,10 @@ def keep_together_if verdict, &block
     end
   end
 
+  def play_icon
```
mojavelinux: I'd rather keep this inline rather than adding a method (even though it's duplicated once).

mojavelinux:
I posed a few questions. If you could answer those, then I think we'll be able to move forward with merging.

lib/asciidoctor-pdf/converter.rb
jessedoyle: Yes it is (see [here](https://github.com/jessedoyle/prawn-icon/blob/97c100edfedd9135b3d00f52a6286a2935abf64e/lib/prawn/icon/interface.rb#L84-L91)). 

`make_icon` is a public method in `Prawn::Icon`'s interface that instantiates a new icon instance without immediately rendering it to the document. 

The method also performs the mapping of FontAwesome 4 keys to FontAwesome 5 keys for compatibility.

lib/asciidoctor-pdf/prawn_ext/extensions.rb
jessedoyle: It's mostly a sanity check, but I believe there may be negative outcomes if we don't exclude it here:

* FontAwesome 5 no longer ships with a base `fa.tff` (or equivalent) font file. Instead the fonts are split into multiple files based on the category (`brands`, `regular` or `solid`). So if we tried to load a base `fa` font, an exception would likely be raised.
* `Prawn::Icon` still accepts FontAwesome 4 specifiers in the format of `fa-{icon-name}` for full backwards compatibility. `Prawn::Icon` contains a compatibility shim and knows how to map these legacy icon specifiers to the corresponding FontAwesone 5 family and definition. 

An icon specifier like **`fa-automobile`** is completely valid, but it maps directly to the **`fas-car`** specifier as seen [here](https://github.com/jessedoyle/prawn-icon/blob/master/data/fonts/fa4/shims.yml#L21). This means that we actually want to load the `fas` (FontAwesome Solid) font instead.

lib/asciidoctor-pdf/prawn_ext/extensions.rb
jessedoyle: Sure!

jessedoyle:
@mojavelinux - I think I've addressed your questions and comments! Please let me know if I can answer any questions.

As my branch already had a merge commit on it, I opted to merge in master as opposed to rebasing. I hope you don't mind!

mojavelinux:
Thanks! I'll have a look tonight.

lib/asciidoctor-pdf/converter.rb
mojavelinux: :+1:

lib/asciidoctor-pdf/prawn_ext/extensions.rb
mojavelinux: If I remember correctly, this block is lazy loading the font. So if we don't lazy load when "fa" is used, then I would expect the processor to crash. That is, unless Prawn::Icon is now loading behind the scenes. I'll need to do some testing.

mojavelinux:
Does prawn-icon follow semantic versioning? If so, we can loosen the dependency to `~> 2.3.0`.

mojavelinux:
It seems like fa-mars and fa-venus are missing from the fa4 shim. I don't know if there are others...I just happened to have those in a test document.

mojavelinux:
Actually, I think I see the problem. It's the other issue I mentioned above that the font set for "fa" is not lazily loaded. I need to figure out how to handle this.

mojavelinux:
Aha. That's why I didn't use make_icon. Because of the way Asciidoctor loads Prawn, that function is not always available. So it's necessary to use the low-level call. I can't exactly remember why this is, other than the fact that Prawn just loads in a very roundabout way.

mojavelinux:
Okay, I think I've got it worked out. Proposal on the way.

mojavelinux:
@jessedoyle I added a commit. See if this still works for the cases you tested.

mojavelinux:
> the fact that Prawn just loads in a very roundabout way.

I want to rephrase this. Technically, the fault lies with Asciidoctor. Asciidoctor can invoke a method on the converter before the converter has been fully initialized. Regardless, we can't always rely on weaved in functions.

jessedoyle:
@mojavelinux - Awesome! I haven't had the chance to test your changes, but I will tonight. Looks good though 👍

jessedoyle:
@mojavelinux - I ran through a few tests as well and it's working well on my end! Thank you very much!

> Does prawn-icon follow semantic versioning? If so, we can loosen the dependency to ~> 2.3.0.

Yes, the gem follows semantic versioning. It would absolutely be okay to loosen the lock to `~> 2.3.0`.
I could push up a commit to this branch if you would like?

mojavelinux:
Thanks for reviewing, @jessedoyle!

About the lock, no need to update. I already set it that way. I was just looking for a confirmation before proceeding with it.

mojavelinux:
@jessedoyle While I was making the changes, I was thinking it might be nice to keep the fact that FontAwesome uses multiple sets abstracted away, even over the long term. This would work as long as the glyphs are mutually exclusive across the three sets. Do you know if that's true? I think it would be really annoying for the user to have to keep track of which FontAwesome set an icon is located in. This is easily the most difficult part of this migration.

jessedoyle:
@mojavelinux - I believe there is some overlap of icon glyphs between the font families.

For example take a look at `address-card`:

* `fas-address-card` - [link](https://fontawesome.com/icons/address-card?style=solid)
* `far-address-card` - [link](https://fontawesome.com/icons/address-card?style=regular)

While the two icons above are similar, they are quite different in style. In FontAwesome 4, this was solved with `*-o`variants ([example](https://fontawesome.com/v4.7.0/icon/address-book-o)).

It is worth noting that in most (possibly all) cases where there is overlap for a given icon, the `solid` variant is considered the default (`fas-*`).

mojavelinux:
Hmm, that's what I was afraid of. Thanks for the clarification.

It's really unfortunate that FontAwesome decided to make such a drastic switch without a good upgrade path. What you've done in prawn-icon definitely smooths out that transition, so I think the solution we have in place now is going to get us where we need to go. Thanks!

(Upgrading FontAwesome in Asciidoctor HTML is going to be a whole other challenge, but that's a problem for me to solve on another day).

mojavelinux:
I think I may pull the calls to Prawn::Icon::FontData.load into a method (which will also allow us to refactor easier). Once I do that, I'll proceed with merging.

mojavelinux:
Done.

PR #940, opened 2018-10-10
Change on reverse.each

TimurTarasenko:
Change on reverse_each because reverse_each loops in reverse order (no
intermediate array created).


mojavelinux:
Good idea. Thanks!

PR #945, opened 2018-10-16
resolves #606 allow running content to be added to toc and title pages

### LINKED ISSUE #606: header on a title and toc pages
Is there a way in theme customization to have header on a title and toc page?

I have the following in a theme:

```
 header:
   font_size: $base_font_size_small
   height: $base_line_height_length * 6.0
   vertical_align: middle
   recto:
       columns: "<30% 40% >30%"
       left:
         content: 'image:logo/logo-1.png[a, scaledwidth="20%"]'
       right:
         content: 'image:logo/logo-2.png[a, scaledwidth="20%"]'
   verso:
       columns: "<30% 40% >30%"
       left:
         content: 'image:logo/logo-1.png[a, scaledwidth="20%"]'
       right:
         content: 'image:logo/logo-2.png[a, scaledwidth="20%"]'
```

habamax:
Theme setup:

```
header:
    startfrom: title|toc

footer:
    startfrom: title|toc
```

https://github.com/asciidoctor/asciidoctor-pdf/issues/606

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -10,6 +10,7 @@ For a detailed view of what has changed, refer to the {uri-repo}/commits/master[
 * drop support for Ruby < 2.3 (and installation will fail for Ruby < 2.1)
 * switch to tilde dependency versions (flexible patch number) instead of ranges
 * upgrade prawn-svg to 0.29.1; resolves numerous SVG rendering issues (#886, #430)
+* allow running content (header and footer) to be enabled on title and toc pages; controlled by running_content_start_at property in theme (#606)
 * allow additional style properties to be set per heading level (#176) (@billybooth)
 * set line spacing for non-AsciiDoc table cells (#296)
 * consider all scripts when looking for leading alpha characters in index (#853)
--- a/docs/theming-guide.adoc
+++ b/docs/theming-guide.adoc
@@ -3149,6 +3149,13 @@ The keys in this category control the arrangement and style of running header an
 |===
 |Key |Value Type |Example
 
+3+|[#key-prefix-running_content]*Key Prefix:* <<key-prefix-running_content,running_content>>
+|start_at
+|title {vbar} toc {vbar} body +
+(default: body)
+|running_content:
+  start_at: toc
+
 3+|[#key-prefix-header]*Key Prefix:* <<key-prefix-header,header>>
 
 |background_color^[1]^
--- a/lib/asciidoctor-pdf/converter.rb
+++ b/lib/asciidoctor-pdf/converter.rb
@@ -206,16 +206,33 @@ class Converter < ::Prawn::Document
     #start_new_page if @ppbook && verso_page?
     start_new_page if @media == 'prepress' && verso_page?
 
-    num_front_matter_pages = (@index.start_page_number = page_number) - 1
+    body_start_page_number = page_number
+
+    if doc.doctype == 'book' || (doc.attr? 'title-page')
+      # NOTE start running content from title or toc, if specified (default: body)
+      if @theme.running_content_start_at == 'title'
+        num_front_matter_pages = 0
+      elsif insert_toc && @theme.running_content_start_at == 'toc'
+        num_front_matter_pages = 1
+      else # body
+        num_front_matter_pages = body_start_page_number - 1
+      end
+    else
+      num_front_matter_pages = body_start_page_number - 1
+    end
+
+    @index.start_page_number = num_front_matter_pages + 1
     font @theme.base_font_family, size: @theme.base_font_size, style: @theme.base_font_style.to_sym
     doc.set_attr 'pdf-anchor', (doc_anchor = derive_anchor_from_id doc.id, 'top')
     add_dest_for_block doc, doc_anchor
+
     unless doc.doctype == 'book' || (doc.attr? 'title-page') || !doc.header? || doc.notitle
       theme_font :heading, level: 1 do
         align = (@theme.heading_h1_align || (doc.doctype == 'book' ? @theme.heading_align : :center) || @base_align).to_sym
         layout_heading doc.doctitle, align: align, level: 1
       end
     end
+
     convert_content_for_block doc
 
     # NOTE delete orphaned page (a page was created but there was no additional content)
@@ -224,7 +241,7 @@ class Converter < ::Prawn::Document
 
     toc_page_nums = insert_toc ? (layout_toc doc, num_toc_levels, toc_page_nums.first, num_front_matter_pages) : []
 
-    if page_count > num_front_matter_pages
+    unless page_count < body_start_page_number
       unless doc.noheader || @theme.header_height.to_f.zero?
         layout_running_content :header, doc, skip: num_front_matter_pages
       end


habamax:
Yep, that was the thing I faced too.

For my usecase though it works as I usually have page numbering only in footer and I start footer after the toc.

And actually this PR does what you say:

> to set num_front_matter_pages to either 0, 1 or the current expression

depending on theme values: for title it is 0 for toc it is 1 and the current expression if nothing was specified. But it does it separately for header and footer.

habamax:
Anyway, I don't speak ruby :), may be I will dig out later how to do proper page numbering.

nattgris:
I don't speak Ruby either :)

Anyway, my point was that in order to not break page numbering, *all* current uses of num_front_matter_pages need to be set to the same value (0, 1 or the current expression). This includes the argument to layout_toc and add_outline. You don't do that in this PR.

If you do that, my use-case (and the page numbering issue) will be fully solved. I have tried it here, if you want I can do a separate PR with it. The only problem is what to call and where to place the theme option ("frontmatter_endbefore: toc|title"? I don't like it).

Unfortunately, if your use case means you also want to start the header and footer on different pages, you will inevitably break page numbering. But maybe it's acceptable because it's only an issue if you have page numbering in both the header *and* the footer *and* start them on different pages. Regardless, num_front_matter_pages must be set to either the number of skipped pages in the footer or the header, depending on where the numbering is placed. Otherwise, the number that you *do* show in footer/header won't match toc/outline.

habamax:
@nattgris could you try it now?

For me if the footer and header starts from the same page (both on title, both on toc) page numbering is ok.

There is still problem when you have them started on different pages, though.

habamax:
image

image

nattgris:
Yes this works for my case. Numbering appears to be OK for all cases where header_startfrom == footer_startfrom. I think this works as the first step.

As discussed before, numbering breaks when headers and footers starts on different pages, which isn't supported at all today anyway.

Not sure if the MAX in your commit has any value. I mean it wouldn't break less if you just picked one of the choices. Actually best choice is to pick the number corresponding to where the numbering is placed. Can we deduce that from the content of the headers/footers? Seems like a workaround though, probably better to fix layout_running_content to generate the correct numbering regardless of skip parameter. Maybe that should be step 2.

habamax:
>Yes this works for my case. Numbering appears to be OK for all cases where header_startfrom == footer_startfrom. I think this works as the first step.

Good to hear it!

I think I myself will not be able to fix numbering if header and footer would start from different pages.

What I think might be a step 1 solution is to have only 1 option both for header and footer, probably `page` option.

```
page:
   headerAndFooterStart: title|toc
```

@mojavelinux do you have any objections/suggestions?

PS
or maybe use `header_startfrom` for both header and footer?

habamax:
So with the #975 it looks like we should apply this only for documents with `book` type.

habamax:
@mojavelinux I have updated pull request to have

1. single option `base_headerfooter_startfrom` 
2. applied only for `Book` doctype

I myself have only 1 issue with it -- name of the option itself.

What about `base_frontmatter_ends_at = title|toc` ?

habamax:
So, it looks like article doctype should also be able to render/not render headerFooter on title page. (`:title-page:` option will create a separate page for a title)

I have updated PR.

mojavelinux:
@habamax Thank you for your work on this and thank you @nattgris for testing and feedback. The hardest thing about changes like this is that we don't break existing documents. I'll do some of my own testing to verify as well.

I'm not convinced the name of the configuration property is right. I think I'd prefer `start_at` (or similar). And if the setting can't be set differently for header and footer, then the category should be running_content. In other words:

```yaml
running_content:
  start_at: toc
```

(The default value is implied to be "body").

mojavelinux:
The theme configuration uses snake case as a naming convention, so we need to follow that regardless of what we call the property.

habamax:
Had to revive this repo :)

I have renamed the property -- did my own tests -- works as expected.

image

@mojavelinux fyi

mojavelinux:
Thanks @habamax.

The `title-page` attribute emulates the title-page behavior of the book doctype, so the two scenarios should share the same logic. In other words, the top of the document looks exactly the same if the doctype is book or the title-page attribute is set.

Also, we have to consider the manpage doctype, so the only two doctypes to check for are book and not book.

habamax:
> Also, we have to consider the manpage doctype, so the only two doctypes to check for are book and not book (which is what your code is doing, so that part is 👍).

Just to clarify -- I still have to check for title-page attr, right?

image

mojavelinux:
That's exactly right.

mojavelinux:
Most excellent. I'll test it out shortly. If everything looks good from my end, we'll precede with merging.

habamax:
Looks good for me -- I had to initialize skip_pages out of the if else block -- otherwise it fails to create a document for article without a title page

habamax:
Anyway article without title-page but with toc attr is broken now.
Default title should be generated first, before toc.

mojavelinux:
For now, feel free to add an exception for "not book & with toc". The underlying problem we'll have to deal with separately. In that case, the TOC should not be on it's own page, but rather directly in the flow of the document.

habamax:
I guess not now :)

Let us wait proper toc handling

mojavelinux:
Well, we don't want Asciidoctor PDF to fail. So we need to at least make sure it generates something sensibly in that case.

If you'd like, you can follow up with a PR to place the TOC correctly in an article. I doubt I'll have time to look at that anytime soon. Perhaps start with submitting an issue so we can track it.

habamax:
To be frank, I had an attempt to do it properly -- having title and then toc after it for an article document type. But I have deleted it a month or so ago -- I wasn't sure 1) if this will be needed, 2) I did first PR "wrongly" as a master not a separate branch, so I couldn't have it in the repo (Well, I could but didn't want to waste a time)

When I did it -- I had a time to spare -- now there is not so much of it. But will try to come up with smth.

mojavelinux:
If all goes well, once we get this next release out, this project is going to gain more autonomy so it can continue to move forward at a faster pace. So good things are in store!

mojavelinux:
I tested it and I agree this works as designed. The setting is effectively ignored for an article. But that should be a non-issue once the TOC is placed correctly.

I'm ready to merge if you are. Could you rebase against master and also add an entry to the CHANGELOG file?

habamax:
>I'm ready to merge if you are. Could you rebase against master and also add an entry to the CHANGELOG file?

If you decipher what rebase against master is. :)

Should I 

1. create a new fork
2. add changes there including changelog -- to be a single commit (?)
3. create a new pull request?

I am not a git(hub) pro so pardon my newb questions

habamax:
I am asking because I don't know how to do history rewrite for a published commits properly

mojavelinux:
In that case, I'll take care of it (when I merge). Just add a commit to update the CHANGELOG.

As general advice, I strongly recommend you learn how to rebase with git because it's a fundamental software development skill nowadays. It's how you reposition your change onto the top of the version history so that you can be sure your changes are based on the last version of the code. There are thousands of tutorials online for how to do it (because git is so prevalent). Here are a few:

* https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase
* https://medium.freecodecamp.org/an-introduction-to-git-merge-and-rebase-what-they-are-and-how-to-use-them-131b863785f
* https://nathanleclaire.com/blog/2014/09/14/dont-be-scared-of-git-rebase/

habamax:
I am not a software developer nowadays, but thank you.

I have a general knowledge of rebase stuff with git but never used it extensively (only to rewrite local history -- to squash several commits)

I will read those docs linked, thanks again.

mojavelinux:
Don't downplay your talents. The fact that you're working on an open source project makes you a software developer by very definition. And I believe the added skill of git rebase will actually save you a lot of time.

mojavelinux:
I've tested this thoroughly and made some updates. I believe it's ready to merge! :tada:

habamax:
Sounds good!

habamax:
Now I can see that history rewrite is not that scary 😨 thing. It looks beautiful

image

mojavelinux:
Indeed, it is a think of beauty.

The best way to think about it is that, in the end, every commit is just a patch. And git is really good at consolidating those patches together into a single patch. Then, you can see the the change very clearly. Neat and tidy.

mojavelinux:
I made one more update to document the new key in the theming guide.

mojavelinux:
Thanks @habamax!

PR #953, opened 2018-11-02
resolves #828 add support for SVG admonition icons

### LINKED ISSUE #828: SVG admonition icons don't work
Using:

Asciidoctor PDF 1.5.0.alpha.15 using  Asciidoctor 1.5.5 [http://asciidoctor.org]
Runtime Environment (ruby 2.3.3p222 (2016-11-21 revision 56859) [x64-mingw32]) (lc:IBM437 fs:Windows-1252 in:- ex:IBM437) 

With:

:icontype: svg

I get:

asciidoctor: WARNING: could not embed admonition icon image: images/icons/warning.svg; image file is an unrecognised format

SVGs work fine as images, can they be allowed for icons as well?

keith-packard:
This checks the 'icontype' and uses prawn-svg to render svg icons for admonitions.

Signed-off-by: Keith Packard <keithp@keithp.com>

### DIFF
--- a/lib/asciidoctor-pdf/converter.rb
+++ b/lib/asciidoctor-pdf/converter.rb
@@ -528,7 +528,6 @@ class Converter < ::Prawn::Document
       label_min_width = label_min_width.to_f
     end
     icons = (node.document.attr? 'icons') ? (node.document.attr 'icons') : false
-    icontype = (node.document.attr? 'icontype') ? (node.document.attr 'icontype') : 'png'
     if icons == 'font' && !(node.attr? 'icon', nil, false)
       icon_data = admonition_icon_data(label_text = type.to_sym)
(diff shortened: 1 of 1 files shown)


mojavelinux:
@keith-packard thanks for doing the legwork on this one! This looks great.

I pushed an additional commit with some improvements. Most notably, I resized the SVG if it doesn't fit in the available space. Have a look and let me know if that change looks good to you. I'll then squash the commits together when I merge.

Could you also revise the commit message to begin with "resolves #828"? Thanks!

keith-packard:
I've updated the commit message in the base patch to prefix it with 'resolves #828'. Thanks for the quick review!

mojavelinux:
Thanks for updating, @keith-packard!

PR #963, opened 2018-11-18
use new logging subsystem if available; otherwise, use shim

mojavelinux:


mojavelinux:
merged as 07ee0c911df9ae8045dc793f39822f6c7dcde0e2

PR #968, opened 2018-11-23
resolves #910 Need to reset chapter and section when starting part

### LINKED ISSUE #910: Need to reset chapter and section when starting part
I have a multipart book, and I have major parts indicated by separate pages, e.g. "Section 3: Library Functions".  I've hacked up an extensions module to layout the library page properly, but I'm finding that I cannot use the running content (header, probably also footer, though I didn't test that) with the {chapter-title} or {section-title}, because what happens is that my otherwise nice clean part separator page now has a running header.  This is ugly, but it gets worse.  The running header now has the last chapter of the previous part (becauseI don't have any new chapters on this page.) 

The solution, after much gnashing of teeth, is to fix asciidoctor-pdf itself -- specifically we need to reset the chapter-title and section-title when starting a new part.  Probably a similar change should be to reset the section-title when starting a new chapter (*if* the chapter is starting on a new page).

With this change, the {chapter-title} evaluates to an empty string on a page containing a new part, but without a chapter.  This is 1000% better than having the bogus and incorrect chapter title that was before.

Here's the diff:

```
diff --git a/lib/asciidoctor-pdf/converter.rb b/lib/asciidoctor-pdf/converter.rb
index 0a66400..638297b 100644
--- a/lib/asciidoctor-pdf/converter.rb
+++ b/lib/asciidoctor-pdf/converter.rb
@@ -2482,6 +2482,8 @@ class Converter < ::Prawn::Document
       if is_book && ((sect_is_part = sect.part?) || sect.chapter?)
         if sect_is_part
           part_start_pages[page_num] ||= (sect.numbered_title formal: true)
+          chapter_start_pages[page_num] = ''
+          section_start_pages[page_num] = ''
         else
           chapter_start_pages[page_num] ||= (sect.numbered_title formal: true)
           if sect.sectname == 'appendix' && !part_start_pages.empty?
```

gdamore:


mojavelinux:
While this change works, the more correct place to clear the value is a few lines down. I made that change in e2713187023fb46f246b92c57cdd9b2835b5faf9.

Issue #937, opened 2018-10-05
Call out number inside code block replaced by `not` glyph


r1m:
Hello,
I'm struggling to render conum inside coderay code blocks. I read almost every glyph issues I could find in this repo but I'm still missing something
![code block](https://user-images.githubusercontent.com/137285/46536687-8a4e0100-c8af-11e8-9eec-b546877f35d4.png)

As you can see conum are ok outside code blocks but not inside.

Using 
```
<asciidoctorj.pdf.version>1.5.0-alpha.16</asciidoctorj.pdf.version>
<asciidoctorj.version>1.5.6</asciidoctorj.version>
<jruby.version>9.1.8.0</jruby.version>
```
And my theme with arial ttf extracted from a Win10 install: 
```yaml
font:
  catalog:
    Arial:
      normal: arial.ttf
      bold: arialbd.ttf
      italic: ariali.ttf
      bold_italic: arialbi.ttf
    # Noto Serif supports Latin, Latin-1 Supplement, Latin Extended-A, Greek, Cyrillic, Vietnamese & an assortment of symbols
    Noto Serif:
      normal: notoserif-regular-subset.ttf
      ...skipped...
    # M+ 1mn supports ASCII and the circled numbers used for conums
    M+ 1mn:
      normal: mplus1mn-regular-ascii-conums.ttf
      ...skipped...
    # M+ 1p supports Latin, Latin-1 Supplement, Latin Extended, Greek, Cyrillic, Vietnamese, Japanese & an assortment of symbols
    # It also provides arrows for ->, <-, => and <= replacements in case these glyphs are missing from font
    M+ 1p Fallback:
      normal: mplus1p-regular-fallback.ttf
     ...skipped...
  fallbacks:
    - M+ 1mn
    - M+ 1p Fallback

conum:
  font_family: M+ 1mn
literal:
  font_family: M+ 1mn
base:
  font_family: Arial
```
Thanks for your help

r1m:
Ok I managed to fix this by using M+1mn font for code
```yaml
code:
  font_family: M+ 1mn
```

mojavelinux:
Indeed, the M+ 1mn font has been prepared to provide the number glyphs. If you use your own font, you have to add these glyphs to the font so that Prawn can find them. The alternative is to use a fallback font that provides them.

r1m:
Well as you can see in my first message, I added the font as fallback but it was not used until I explicitly use m+1mn for conum and code.

[...]

PR #970, opened 2018-11-25
Clears thing about glyph missing in default fonts

r1m:
Add some information about missing glyph with default font settings 
Based on the discussion in #937


mojavelinux:
Thanks!

I'll also add that fallback we talked about so that it isn't necessary to set both literal and code.

PR #975, opened 2018-11-27
resolves #105 don't add title page and show doctitle above content when doctype is article

### LINKED ISSUE #105: Prevent Page break after title
When printing a simple 1 page piece of text to PDF, how can I prevent the page break after the document title?

mojavelinux:
resolves #105

### DIFF
--- a/lib/asciidoctor-pdf/converter.rb
+++ b/lib/asciidoctor-pdf/converter.rb
@@ -188,7 +188,7 @@ class Converter < ::Prawn::Document
     end if respond_to? :on_page_create
 
     layout_cover_page :front, doc
-    layout_title_page doc
+    layout_title_page doc if doc.doctype == 'book' || (doc.attr? 'title-page')
 
     # NOTE a new page will already be started if the cover image is a PDF
     start_new_page unless page_is_empty?
@@ -210,6 +210,12 @@ class Converter < ::Prawn::Document
     font @theme.base_font_family, size: @theme.base_font_size, style: @theme.base_font_style.to_sym
     doc.set_attr 'pdf-anchor', (doc_anchor = derive_anchor_from_id doc.id, 'top')
     add_dest_for_block doc, doc_anchor
+    unless doc.doctype == 'book' || (doc.attr? 'title-page')
+      theme_font :heading, level: 1 do
+        align = (@theme.heading_h1_align || (doc.doctype == 'book' ? @theme.heading_align : :center) || @base_align).to_sym
+        layout_heading doc.doctitle, align: align, level: 1
+      end
+    end
     convert_content_for_block doc
 
     # NOTE delete orphaned page (a page was created but there was no additional content)
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -10,6 +10,7 @@ For a detailed view of what has changed, refer to the {uri-repo}/commits/master[
 * allow additional style properties to be set per heading level (#176) (@billybooth)
 * set line spacing for non-AsciiDoc table cells (#296)
 * consider all scripts when looking for leading alpha characters in index (#853)
+* don't create title page for article doctype unless title-page attribute is set (#105)
 * remove unprocessed passthroughs in literal cells
 * apply font style from theme to formatted text description list term (#854)
 * prevent tempfile for remote image from being deleted before it's used (#947)
--- a/data/themes/default-theme.yml
+++ b/data/themes/default-theme.yml
@@ -80,7 +80,7 @@ heading:
   font_color: $base_font_color
   font_family: $base_font_family
   font_style: bold
-  # h1 is used for part titles (book doctype only)
+  # h1 is used for part titles (book doctype) or the doctitle (article doctype)
   h1_font_size: floor($base_font_size * 2.6)
   # h2 is used for chapter titles (book doctype only)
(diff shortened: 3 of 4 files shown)


habamax:
Not sure where to report -- I wanted to test this feature and git cloned the repo.

I get the errors:

```
"C:\prg\Ruby26-x64\lib\ruby\2.6.0\tempfile.rb" [unix] 354L, 11484C
(1 of 30): in `<top (required)>': superclass mismatch for class Tempfile (TypeError)
```

image


About 2 months ago, I did the same thing and it worked just fine...

mojavelinux:
Looks like an incompatibility with Ruby 2.6. Can you try Ruby 2.5?

habamax:
I will try -- rubyinstaller is hosted on a rather slow server...

Anyway current gem version of asciidoctor-pdf works on ruby 2.6

habamax:
Ruby 2.5 has the error too:

image

Opened an issue #996

mojavelinux:
Let's continue the discussion in #996 as this problem doesn't pertain to this PR.

habamax:
I have tested it and without a `:toc:` it looks good.

With a `:toc:` I have a separate first toc page and then the rest of the document:

1. toc
2. doctitle + doc contents


BTW, how can I control the appearance of the doc title in that case? Like alignment, size, color, etc?

mojavelinux:
Currently the doctitle is controlled by the h1 category. There is a proposal to make this theming more advanced. See https://github.com/asciidoctor/asciidoctor-pdf/issues/988

mojavelinux:
From  the default theme:

> # h1 is used for part titles (book doctype) or the doctitle (article doctype)

habamax:
I have found the issue(?) though:

`:notitle:` option is not respected -- I think doctitle shouldn't be rendered if it is set.

mojavelinux:
You are correct. Can you file a separate issue?

habamax:
Sure, done.

lib/asciidoctor-pdf/converter.rb
```diff
@@ -188,7 +188,7 @@ def convert_document doc
     end if respond_to? :on_page_create
 
     layout_cover_page :front, doc
-    layout_title_page doc
+    layout_title_page doc if doc.doctype == 'book' || (doc.attr? 'title-page')
```
habamax: I would add this condition to layout_title_page function itself.
There is even FIXME comment there.

  # FIXME only create title page if doctype=book!
  def layout_title_page doc
    # Do not create title page 
    # 1. if doctype!=book
    # 2. there is no :title-page: attribute
    # 3. there is :notitle: attribute
    # 4. there is no header in the document (= header)
    if doc.doctype != 'book' && !(doc.attr? 'title-page') || doc.notitle || !doc.header?
      return
    end
...

lib/asciidoctor-pdf/converter.rb
```diff
@@ -210,6 +210,12 @@ def convert_document doc
     font @theme.base_font_family, size: @theme.base_font_size, style: @theme.base_font_style.to_sym
     doc.set_attr 'pdf-anchor', (doc_anchor = derive_anchor_from_id doc.id, 'top')
     add_dest_for_block doc, doc_anchor
+    unless doc.doctype == 'book' || (doc.attr? 'title-page')
```
habamax: Adding ` || doc.notitle` will solve the issue #998 

    unless doc.doctype == 'book' || (doc.attr? 'title-page') || doc.notitle

Did it on my fork, works for me (not sure how to properly separate stuff, waiting for my PR to be included/rejected by you and then I will probably be able to properly use feature branches)

lib/asciidoctor-pdf/converter.rb
```diff
@@ -210,6 +210,12 @@ def convert_document doc
...
+      theme_font :heading, level: 1 do
+        align = (@theme.heading_h1_align || (doc.doctype == 'book' ? @theme.heading_align : :center) || @base_align).to_sym
+        layout_heading doc.doctitle, align: align, level: 1
+      end
+    end
```
habamax: And if you move whole block for the article heading generation a bit up, like:

image

so the toc creation would be after heading generation (the same as for book type) then we can have the following:


image

Now if there was an easy way to calculate vertical size of the heading (that might include subtitle and autours and revision date in the future) and shift toc down to that size...

ogmios-voice: Pull request #986 already solves this (see also #233).
It has conflicts now, but I will do the merge after finishing #988.

lib/asciidoctor-pdf/converter.rb
mojavelinux: This will be handled by the fix for #998. I've proposed a PR.

lib/asciidoctor-pdf/converter.rb
mojavelinux: Let's address this in #988 since this PR has already been closed.

lib/asciidoctor-pdf/converter.rb
mojavelinux: I'd rather keep it as it is. Whether layout_title_page is called depends on the doctype or the title-page attribute override (which can easily be set in an extension). The rest of the logic should be handled by the method itself.

There was just a time when we didn't have control over it, so the comment is actually outdated now.

PR #986, opened 2018-12-26
resolves #233 TOC placed directly after title, no page break after TOC when doctype is article

### LINKED ISSUE #233: Can't control PDF TOC output location

ogmios-voice:
Note: It is based on pull request #985.

### DIFF
--- a/lib/asciidoctor-pdf/converter.rb
+++ b/lib/asciidoctor-pdf/converter.rb
@@ -187,8 +187,7 @@ class Converter < ::Prawn::Document
       end
     end if respond_to? :on_page_create
 
-    layout_cover_page :front, doc
-    layout_title_page doc
+    layout_title doc
 
     # NOTE a new page will already be started if the cover image is a PDF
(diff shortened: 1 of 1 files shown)


mojavelinux:
Thanks for getting this started! This PR has now been superseded by #1029. I drew ideas from this PR, but found many additional cases that still needed to be handled. To prevent regressions, I added tests for many of those cases (working within the limits of the current assertion model).

PR #1010, opened 2019-03-22
refs #73 refs #85. Append footnotes at document end

bcourtine:
Footnotes are not inlined anymore. They are appened at document end, with bi-directional links between document and notes.

It is not perfect (footnotes are not at page end), but I think it is better than the current situation.

### DIFF
--- a/lib/asciidoctor-pdf/converter.rb
+++ b/lib/asciidoctor-pdf/converter.rb
@@ -111,6 +111,7 @@ class Converter < ::Prawn::Document
     outfilesuffix '.pdf'
     @list_numbers = []
     @list_bullets = []
+    @footnotes = []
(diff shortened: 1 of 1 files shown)


mojavelinux:
Thank you for getting the ball rolling on this, @bcourtine! I have proposed a revised implementation based on this work in PR #1016. Please give it a try and let me know how it work out.

The main differences are that a) the footnotes added in a chapter are placed at the end of that chapter when the doctype is book, b) it uses the same convention for generating IDs as is used in core, c) changes the heading to a block title, adding it only if footnotes-title is set, and c) adds rudimentary theme control.

mojavelinux:
I'm closing this in favor of #1016, which builds on this proposal.

PR #1016, opened 2019-04-07
resolves #85 place footnotes at end of chapter (book) or document (article)

### LINKED ISSUE #85: Footnotes are not positioned correctly

mojavelinux:
- place list of footnotes at end of chapter (when doctype is book) or document (when doctype is article)
- add footnotes category to theme to style footnote list
- add additional footnotes to example documents
- honor footnotes-title attribute to control title of footnote list (unset by default)

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -12,6 +12,7 @@ For a detailed view of what has changed, refer to the {uri-repo}/commits/master[
 * upgrade prawn-svg to 0.29.1; resolves numerous SVG rendering issues (#886, #430)
 * allow running content (header and footer) to be enabled on title and toc pages; controlled by running_content_start_at property in theme (#606)
 * upgrade to FontAwesome 5; introduce the fas, far, and fab icon sets, now preferred over fa; drop support for octicons (#891) (@jessedoyle)
+* place footnotes at end of chapters in books or end of document otherwise (#85) (@bcourtine)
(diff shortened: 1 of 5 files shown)


bcourtine:
I tested your PR and it looks good to me.

bcourtine:
I thought of an evolution, but don't know how to implement it: if `footnotes-title` is defined, corresponding page could be added to the TOC (maybe with another option `footnotes-toc`).

mojavelinux:
I don't think that's the right use of a section heading. Footnotes are not a discrete section or appendix. They are an addendum to a section (chapter for book or document for article). Therefore, I conclude that they map closer to a block (as I have done it) than a section.

bcourtine:
Seems legit with your enhancements. I was still thinking about my first version where all footnotes were in a unique end section (sometimes several pages long).

mojavelinux:
Indeed, if you have chapters, the footnotes should be broken up by chapter. They would only appear at the very end if you have no chapters or it's an article. This is consistent with how I've seen footnotes be handled in ebooks.

gAmUssA:
@mojavelinux which version of asciidoctorj we have this feature?

mojavelinux:
The version of AsciidoctorJ PDF matches the version of Asciidoctor PDF.
Just consult the changelog (search for 85) and you will know.

>

PR #1023, opened 2019-04-11
Add unicode support to index terms

zyuhel:
Now index terms for non latin characters are treated as other, this makes impossible to use index for languages like cyrillic. Extending alpha regex to use unicode will fix it

### DIFF
--- a/lib/asciidoctor-pdf/index_catalog.rb
+++ b/lib/asciidoctor-pdf/index_catalog.rb
@@ -1,6 +1,6 @@
 module Asciidoctor; module Pdf
   class IndexCatalog
-    LeadingAlphaRx = /^\p{L}/
+    LeadingAlphaRx = /^\p{Alpha}/
 
     attr_accessor :start_page_number
 
--- a/lib/asciidoctor-pdf/index_catalog.rb
+++ b/lib/asciidoctor-pdf/index_catalog.rb
@@ -1,6 +1,6 @@
 module Asciidoctor; module Pdf
   class IndexCatalog
-    LeadingAlphaRx = /^[[:alpha:]]/u
+    LeadingAlphaRx = /^\p{L}/
 
     attr_accessor :start_page_number
 
--- a/lib/asciidoctor-pdf/index_catalog.rb
+++ b/lib/asciidoctor-pdf/index_catalog.rb
@@ -1,6 +1,6 @@
 module Asciidoctor; module Pdf
   class IndexCatalog
-    LeadingAlphaRx = /^[[:alpha:]]/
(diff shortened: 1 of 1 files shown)


lib/asciidoctor-pdf/index_catalog.rb
```diff
@@ -1,6 +1,6 @@
 module Asciidoctor; module Pdf
   class IndexCatalog
-    LeadingAlphaRx = /^[[:alpha:]]/
+    LeadingAlphaRx = /^[[:alpha:]]/u
```
mojavelinux: I would prefer that you use the following: `\p{Alpha}`. This is consistent with what we have done in core and elsewhere.

zyuhel: When it should be `\p{L}` not `\p{Alpha}` for unicode support as far as i know

mojavelinux: \p{Alpha} in Ruby includes all Unicode. This is documented specifically in the language documentation.

zyuhel: rechecked, \p{Alhpa} is fine with cyrillic in ruby.

mojavelinux: In fact, `[[:alpha:]]` on Ruby is already Unicode aware.

```
/[[:alpha:]]/.match? 'é'
# => true
```

The only thing that doesn't support unicode are character ranges:

```
/[a-z]/.match? 'é'
```

mojavelinux: But I still like the idea of using `\p{Alpha}` instead of `[[:alpha:]]` because it's more clear that it's modern and thus Unicode aware. (Very old versions of Ruby in fact didn't support Unicode in `[[:alpha:]]`).

mojavelinux: Can you give me sample words that your testing (until we have a test suite).

zyuhel: here is cyrillic words i'm using together that similar latin ones

Абв
Abc
Бвг
Модульный
Modular

zyuhel: yes it seems that there is no issue with current master  :(
i rechecked, i had v1.5.0 Alpha 16 locally installed, instead of regexp it uses ord with character range.

mojavelinux: Excellent. We'll proceed with this regardless for consistency.

Issue #994, opened 2019-01-16
side margins for running content should be configurable


mkvonarx:
Currently, the header/footer border always uses the page_margin for it's left/right position.

When changing the left/right padding of the header/footer, the header/footer border should use these instead of the page_margin to calculate the left/right position of the header/footer border lines.

beatchristen:
I had a similar issue in the other direction: border should extend all the way to the edge of the page.
My suggested fix is in https://github.com/systarch/asciidoctor-pdf/tree/refactor-header-footer.

The integration test needs some cleaning first before I will submit a PR.

mojavelinux:
> I had a similar issue in the other direction: border should extend all the way to the edge of the page.

If you set a background color (even if white), both the background and border extend to the edge of the page. I prefer this strategy. One modification we could consider is to use this logic even if the background color is set to transparent.

> Currently, the header/footer border always uses the page_margin for it's left/right position.

That's the starting position. You can use negative values to pull the content outwards. However, I noticed that the border and background color don't honor the negative values, but I'm not sure what the right behavior would be.

What we probably need is a running content margin. Right now the margin and padding are coupled. That's the change I would focus on making.

mojavelinux:
> What we probably need is a running content margin

By default this would match the page margin. If we were to allow this property to be set, I think it would solve both issues.

mojavelinux:
I'll update the issue accordingly.

mojavelinux:
The tricky part about this issue is that the page side margin can differ by page. When media=prepress, it can differ for recto and verso pages. When the page layout is changed, it can differ for that page. So we'll need some way to be able to refer to that value with a keyword.

Here's what I propose:

* The padding is computed relative to the margin (as one would expect).
* By default, the margin is `[0, initial]`, where `initial` represents the current side margin of the page on either side.
* The content can be aligned with the content of the page using a value of `auto`. So if the margin is 0, the padding can be set to `[0, auto]` to align the running content with the page content column.
* Setting the background color and image will no longer implicitly set the side margin to 0.

For example, f you want the border to span the width of the page, but the content to be restricted to the width of the page content, use:

```
footer:
  border_width: 0.25
  border_color: dddddd
  margin: 0
  padding: [3, auto]
  ...
```

The `auto` keyword will assign a padding that's the difference of the page side margin and the running content side margin so that the content aligns with the page content.

To give another example, you can allow the border to extend on one side:

```
footer:
  border_width: 0.25
[...]

PR #1024, opened 2019-04-14
visual regression tests

beatchristen:
Hi @mojavelinux 
it's currently not the configurable solution you mentioned in #994 with side margin values.
But you might be interested in the visual regression test it comes with.

caveat emptor:
- I have no idea what the compare_channel result `diff_metric` will spit out on different platforms.
- fixtures in `spec/integration/fixtures/layout_periphery_boundingbox_spec`: is this too rigid?
- should a fixture image be created automatically if it's not present?
- if the compare fails: How should it present the results?


mojavelinux:
First of all, I love that you've added a test suite. That's so desperately needed in Asciidoctor PDF. As chance would have it, you caught me while I was also preparing a test suite, now staged in #1025. As soon as that's merged, can you rebase onto it and use that foundation? (which is also based on RSpec).

I have to believe there's a simpler way to install Imagemagick in Travis. I would really like to avoid maintaining a custom script. Is it possible to use apt-get to install it instead? And outside of Travis, it should be up to the developer to know how to install it. We have to draw the line somewhere as to what we are willing to maintain.

I definitely like the idea of using visual testing. There's some stuff we absolutely cannot test by just looking at the PDF objects alone (or it would be very fragile). However, we probably want to only use these when there is no other way to test because it can slow down the test suite (and thus development) considerably.

beatchristen:
> First of all, I love that you've added a test suite. That's so desperately needed in Asciidoctor PDF. As chance would have it, you caught me while I was also preparing a test suite, now staged in #1025. As soon as that's merged, can you rebase onto it and use that foundation? (which is also based on RSpec).

I have seen your fork with the rspec work, I will rebase versus that one to get some suggestions in as well.

> 
> I have to believe there's a simpler way to install Imagemagick in Travis. I would really like to avoid maintaining a custom script. Is it possible to use apt-get to install it instead? And outside of Travis, it should be up to the developer to know how to install it. We have to draw the line somewhere as to what we are willing to maintain.

The imagemagick part is crazyiness, absolutely agree.
I will kick that one out as it's too brittle and differs even across linuxens, probably due to ghostscript having a different version.
As there's no way to get the generated binary out from travis (or is there?), I will pursue two different strategies:
- poppler has an alternative implementation for PDF and it looks like it is available for win, lin and mac. most platforms should have a binary gem available, removing the need to build from source.
- the reference image should become a reference PDF which is rendered by the same mechanism. This should produce much more stable results when comparing.

> 
> I definitely like the idea of using visual testing. There's some stuff we absolutely cannot test by just looking at the PDF objects alone (or it would be very fragile). However, we probably want to only use these when there is no other way to test because it can slow down the test suite (and thus development) considerably.

The visual regression testing probably makes sense to run only when you are about to produce a release.
OTOH, it's the installation of imagemagick that took 6-8 minutes, the test itself is reasonably fast with approximately 300-500ms and would also be a nice "feature gallery" when used on single item configurations

beatchristen:
I have removed some no longer needed items like env variable IMAGEMAGICK and the previous png fixture.
Apart from that, out of ideas what you would expect beyond the current implementation.

Would you be able to share the theme variables you had in mind when in #994 you said:
> If you set a background color (even if white), both the background and border extend to the edge of the page. I prefer this strategy.

Then I'd be able to adapt the test and it wouldn't clash with the #1027

beatchristen:
Hmm, so once the module has been prepended it lives on and messes up the tests later down the test road.
Any idea how to 
1. ... remove the registered module 'LayoutPeripheryBoundingBoxExtension'?
2. ... tell rspec to run this test in isolation?
3. ... inject the theme variables from code without a module?

beatchristen:
setting the header_border_width and footer_border_width to 0 did the trick.
So now I have removed the original refactoring method from the converter and it's all down to theme configuration.

@mojavelinux Any ideas how to improve loading the theme variables for tests?

mojavelinux:
> Any ideas how to improve loading the theme variables for tests?

I need to add support for this to the `to_pdf` helper, or perhaps some other way of specifying a theme override. I'll add that then post when it's ready.

mojavelinux:
It's now possible to pass a Hash of overrides to the `to_pdf` helper using the `:theme_overrides` option. Here's an example: https://github.com/asciidoctor/asciidoctor-pdf/blob/422a2cf6470f2c3c0de4950c5c4070ef1a139ac6/spec/document_title_spec.rb#L47

.gitignore
```diff
@@ -8,3 +8,4 @@
 /.ruby-gemset
 /.ruby-version
 /.yardoc/
+/spec/output/
```
mojavelinux: This folder should get cleaned up automatically by the test suite.

.travis.yml
```diff
@@ -1,9 +1,15 @@
...
+sudo: required
+before_install:
+  - source .travis/before_install_$TRAVIS_OS_NAME.sh
 git:
   # use depth 2 just in case two refs get pushed at once (like a tag)
   depth: 2
 language: ruby
-cache: bundler
+cache:
+  bundler: true
```
mojavelinux: I don't want to cache bundler because I want to verify installation works from scratch. We have had problems in the past an installation time and we need to verify it works.

Gemfile
```diff
@@ -10,6 +10,11 @@ group :examples do
   #gem 'unicode' if (Gem::Version.new RUBY_VERSION) < (Gem::Version.new '2.4.0')
 end
 
+group :test, :development do
+  gem 'rspec'
```
mojavelinux: For the time being, development dependencies should be listed in the gemspec (where you will find rspec already listed).

spec/spec_helper.rb
```diff
@@ -1,14 +1,13 @@
 require 'asciidoctor-pdf'
+require 'chunky_png'
+require 'fileutils'
```
mojavelinux: fileutils is already pulled in by Asciidoctor PDF.

spec/integration/layout_periphery_boundingbox_spec.rb
```diff
@@ -0,0 +1,39 @@
+require 'spec_helper'
+
+describe 'when configuring the header and footer background colors' do
+
+  class PeripheryExtensionTest < ::Asciidoctor::Pdf::Converter
+
+    def build_pdf_options doc, theme
```
mojavelinux: It's now possible to pass these overrides to the `to_pdf` helper using the `:theme_overrides` option.

spec/integration/layout_periphery_boundingbox_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+      super doc, theme
+    end
+
+  end
+
+  it "should draw the configured color across the whole periphery region", :integration => true do |test|
+
+    PeripheryExtensionTest.register_for 'pdf'
+
+    Asciidoctor.convert "Hello world",
```
mojavelinux: Please use the `to_pdf` helper.

.travis.yml
```diff
@@ -1,9 +1,15 @@
 dist: trusty
+sudo: required
+before_install:
+  - source .travis/before_install_$TRAVIS_OS_NAME.sh
```
mojavelinux: Please use the following syntax:

```yml
addons:
  apt:
    packages:
    - poppler-utils
```

We should not be changing the sudo setting.

.travis/before_install_linux.sh
```diff
@@ -0,0 +1,8 @@
+set -euox pipefail
```
mojavelinux: We don't need this script as Travis CI can handle this for Linux.

.travis/before_install_osx.sh
```diff
@@ -0,0 +1,3 @@
+
```
mojavelinux: I'm not really interested in an macOS script. We test on Linux only and I don't want to change that in this PR.

spec/integration/layout_periphery_boundingbox_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+      theme.header_height = 160
+      theme.footer_height = 160
+      theme.page_margin = [ 160, 48, 160, 48]
+
+      super doc, theme
+    end
+
+  end
+
+  it "should draw the configured color across the whole periphery region", :integration => true do |test|
```
mojavelinux: Please use the modern Hash syntax, `integration: true`. Also, I don't think the block need an argument since I don't see it used anywhere.

spec/spec_helper.rb
```diff
@@ -1,14 +1,13 @@
...
 require 'pathname'
 require 'pdf/inspector'
 
 RSpec.configure do |config|
   config.before :suite do
-    FileUtils.mkdir_p output_dir
```
mojavelinux: Please leave the before and after as they were. The output folder should get cleaned up.

spec/spec_helper.rb
```diff
@@ -48,4 +52,45 @@ def with_memory_logger level = nil
...
+    end
+
+    diff.length
+  end
+
+end
+
+RSpec::Matchers.define(:match_fixture) do |reference_filename|
+  match do |actual_path|
+    reference_path = File.join __dir__, "integration", "fixtures", testname, reference_filename
```
mojavelinux: Please use the folder spec/fixtures/integration/...

spec/spec_helper.rb
```diff
@@ -48,4 +52,45 @@ def with_memory_logger level = nil
...
+      images.last.rect(x.min, y.min, x.max, y.max, ChunkyPNG::Color.rgb(0,255,0))
+      images.last.save(difference)
+    end
+
+    diff.length
+  end
+
+end
+
+RSpec::Matchers.define(:match_fixture) do |reference_filename|
```
mojavelinux: match_fixture seems too generic of a name. It should be something like "has_visual_match" or "visually_matches".

spec/spec_helper.rb
```diff
@@ -19,6 +18,11 @@ def fixture_file path
     File.join fixtures_dir, path
   end
 
+  def testname
+    metadata = method_missing(:class).metadata
```
mojavelinux: This could be stored using the following:

```rb
config.before :each do |example|
  @example_group_name = File.basename (example.instance_variable_get :@metadata)[:file_path], '.rb'
end
```

mojavelinux: In fact, it's even simpler:

```rb
RSpec.configure do |config|
  config.expose_current_running_example_as :example

  config.before :each do
    @example_group_name = File.basename example.file_path, '.rb'
  end
end
```

spec/integration/layout_periphery_boundingbox_spec.rb
mojavelinux: If we're going to mark tests as integration, then they don't need to be in a separate folder.

spec/spec_helper.rb
```diff
@@ -48,4 +52,45 @@ def with_memory_logger level = nil
...
+    reference_path = File.join __dir__, "integration", "fixtures", testname, reference_filename
+    images_output_dir = output_file File.join "integration", testname
+    FileUtils.mkdir_p images_output_dir
+    output_basename = File.join images_output_dir, File.basename(actual_path, '.pdf')
+    %x( pdftocairo -png "#{actual_path}" "#{output_basename}-actual" )
+    %x( pdftocairo -png "#{reference_path}" "#{output_basename}-reference" )
+
+    # find the generated pairs to compare
+    pixelcount = 0
+    Dir["#{output_basename}-reference-*.png"].each do |reference_page_filename|
```
mojavelinux: The convention in Asciidoctor is to define interpolated strings using `%()` instead of `""`.

spec/spec_helper.rb
```diff
@@ -48,4 +52,45 @@ def with_memory_logger level = nil
...
+end
+
+RSpec::Matchers.define(:match_fixture) do |reference_filename|
+  match do |actual_path|
+    reference_path = File.join __dir__, "integration", "fixtures", testname, reference_filename
+    images_output_dir = output_file File.join "integration", testname
+    FileUtils.mkdir_p images_output_dir
+    output_basename = File.join images_output_dir, File.basename(actual_path, '.pdf')
+    %x( pdftocairo -png "#{actual_path}" "#{output_basename}-actual" )
+    %x( pdftocairo -png "#{reference_path}" "#{output_basename}-reference" )
```
mojavelinux: Please remove the spaces around the command.

.travis.yml
beatchristen: Rereading the manual at https://docs.travis-ci.com/user/caching/ makes me believe my change would have been a noop, as it inspects where the gems go with the --path and caches those.
cache: { bundler: true } is only required if there are other directories in need of caching, so let's leave it as is.

spec/integration/layout_periphery_boundingbox_spec.rb
beatchristen: This one turns out to be tricky.
the to_pdf helper is either denying writing outside the spec/fixture or then doesn't write and returns the analyze oder reader response.
Maybe split this into to_pdf_reader / to_analyzer / to_pdf_file?

spec/spec_helper.rb
beatchristen: I notice this section is now gone. Intention was to have the visual output in spec/output/integration/... in case of differences.

spec/integration/layout_periphery_boundingbox_spec.rb
mojavelinux: I can change the implementation of to_pdf to handle any additional requirements.

However, it should still be writing somewhere inside the spec/output directory. That keeps the tests self-contained.

spec/spec_helper.rb
beatchristen: I have removed the testname-based structure in the fixtures section as it's currently just one.
There's also the need to point to the proper path in case the test fails, and at that time, the :example doesn't contain the name yet or not anymore.

spec/integration/layout_periphery_boundingbox_spec.rb
beatchristen: `spec` as jail root should suffice to read from fixtures and write to output.

.travis.yml
mojavelinux: I just don't trust the cache, so I don't want it to be anywhere in the CI config.

mojavelinux:
> Any ideas how to improve loading the theme variables for tests?

the to_pdf function now allows theme variables to be overridden. Consult the test suite for examples of the theme_overrides key.

mojavelinux:
This is really fantastic. Thanks! I pushed some cleanups. I'm just waiting on CI to give us the green light. Then I'll proceed with merging.

mojavelinux:
Once this is merged, I'll make a visual regression test for the two examples in the repository so we don't need the extra call to asciidoctor-pdf in the build.

mojavelinux:
Green bar! I'll merge shortly.

PR #1026, opened 2019-04-17
resolves #490 support page layout per page

### LINKED ISSUE #490: Support page layout (landscape/portrait) per page
Prawn supports different page layout per page. (through start_new_page and Document.new).
Asciidoctor-pdf could support toggle-able page layout for next page to be started.

I hacked abit on the source to experiment a bit, but did not find any clean solution.
I replaced all calls to start_new_page with a super method that takes the current node and retrieves
the :page-layout: attribute from it. But this probably would not work if the start_new_page is called in a block without this attribute set. This attribute should probably work similar to the :numbered: (sectnum?) toggling. Thus, it would need to be present in all blocks? 

Given the following ADoc source:

```
= Snickers: twist
Test Author Of the Year

== First text page
This i cannot get landscape mode...

:page-layout: landscape
<<<

== Landscape page 1
Content 1

:page-layout: portrait
<<<

== Portrait page 1
Content 2

:page-layout: landscape
<<<

== Landscape page 2
Content 3
```

I was able to produce this PDF:
[layout.pdf](https://github.com/asciidoctor/asciidoctor-pdf/files/400383/layout.pdf)

I'm looking for both discussion and suggestions to how this could be implemented (if of course this is at all anything you desire to be part of Asciidoctor-pdf). I'd be more than happy to work on it further.

resort-diaper:
This resolves part of #490, which is the ability to use specify the page layout use a block attribute on a page break as noted in https://github.com/asciidoctor/asciidoctor-pdf/issues/490#issuecomment-237390736. 

There is extra work required in being able to set the page margins and page size as currently these are assumed to be consistent across the entire document. 

The key limitation of this PR is that the headers and footer are not resized when the layout is changed.

### DIFF
--- a/lib/asciidoctor-pdf/converter.rb
+++ b/lib/asciidoctor-pdf/converter.rb
@@ -1889,7 +1889,14 @@ class Converter < ::Prawn::Document
 
   # NOTE to insert sequential page breaks, you must put {nbsp} between page breaks
   def convert_page_break node
-    advance_page unless at_page_top?
+    opts = {}
+    # Detect if a page layout change was a role of the page break.
+    page_layout = resolve_layout_from_node node
+    opts = {
+              layout: (page_layout || page.layout),
+           }
+
+    advance_page opts unless at_page_top?
   end
 
   def convert_index_section node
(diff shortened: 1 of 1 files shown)


mojavelinux:
Thanks! Merged as c5589cd. I simplified the code slightly and added some tests.

PR #1027, opened 2019-04-17
resolves #1036 resize running content per page layouts (PR #1027)

### LINKED ISSUE #1036: Footer page number is badly aligned when [.landscape] is used
There is a new possibility to set page orientation #490 (which is really handy)

The issue is in footer to be rendered in unexpected way.

Having in the theme:
```
footer:
  font_size: $base_font_size_small
  height: $base_line_height_length * 3.0
  line_height: 1
  padding: [$base_line_height_length / 2, 1, 0, 1]
  vertical_align: top
  recto:
    right:
      content: '{page-number}'
  verso:
    right:
      content: '{page-number}'

```

and the file contents:

```
= Test
:experimental:
:icons: font
:autofit-option:
:source-highlighter: rouge
:rouge-style: github
:revdate: 2018-12-07
:imagesdir: images
:doctype: article
:toc:

== test 1
// 2019-04-22

Ut sed lacus id mauris mattis lobortis. Aliquam ut nisl eu neque viverra
pretium. Interdum et malesuada fames ac ante ipsum primis in faucibus. Etiam


[.landscape]
<<<
== hello
// 2019-04-16 12:26:58

Lorem ipsum dolor sit amet, consectetur adipiscing elit.
Maecenas feugiat fermentum pretium. Cras eu dolor imperdiet justo mattis
pulvinar. Cras nec lectus ligula. Proin elementum luctus elit, a tincidunt quam
facilisis non. Nunc quis mauris non turpis finibus luctus. Maecenas ante
sapien, sagittis quis accumsan in, feugiat quis sem. Praesent et auctor libero.

[.portrait]
<<<
== test

hello
```

It gets rendered like this (see the page number):

image

resort-diaper:
This PR expands on PR #1026 to address issue #490 to provide resizing of page margins when different page layouts and pages sizes are used within a single document.

1. It is first determined all the different page layout and size combinations used throughout the document and the appropriate page margins.
2. For each layout and size combination calculate the page margins and store these values.
3. As each page is created set the appropriate page margin from the pre-calculated values.

Note: This PR does not include the ability to set different page margins for each page layouts or sizes in the theme.


mojavelinux:
First, good job getting this going. I have to admit that the amount of code changed makes me nervous, but I also recognize it may be a necessary evil.

Now that we have a test suite in place, every pull request will be required to have tests to verify the behavior. Can you add tests to cover these scenarios?

Testing margins is still a bit tricky since the test suite doesn't yet provide good helpers for this. You can see https://github.com/asciidoctor/asciidoctor-pdf/blob/master/spec/page_spec.rb for what I've done so far. We may need to go deeper and evaluate the PDF objects. Or, we might be able to modify `to_pdf` (or similar) to return the Prawn document so we can look at the state before it is written to PDF.

resort-diaper:
It's probably not the most elegant piece of work ever produced. I will have a look at adding some tests.

mojavelinux:
> It's probably not the most elegant piece of work ever produced.

I wasn't criticizing the code, but rather lamenting on the amount of logic that is necessary to implement this change. In other words, it has quite a bit of ripple.

resort-diaper:
I've added a pretty simple test using the body text and the page number locations to ensure that they are the same distances from the page borders for both portrait and landscape layouts. 

My understanding of the code base is not that great so I am not sure how to modify `to_pdf` (or create a similar method) to return a Prawn document.

mojavelinux:
> I am not sure how to modify `to_pdf` (or create a similar method) to return a Prawn document.

That has been on my list to do, so I'll go ahead and apply that change to master.

mojavelinux:
You can now return the Prawn::Document in the test using:

```
prawn = to_pdf 'text', analyze: :document
```

spec/page_spec.rb
```diff
@@ -124,5 +124,32 @@
...
+      [.landscape]
+      <<<
+
+      landscape-1
+
+      <<<
+
+      landscape-2
+      EOS
+      (expect pdf.positions[0]).to eql [48.24, 793.926]
```
mojavelinux: This is better done using `pdf.find_text`, then look at the x/y position of the results.

mojavelinux:
I'm going to build on this PR with a slightly cleaner proposal.

mojavelinux:
Let me know what you think.

mojavelinux:
I'd like to move forward with this PR. The one thing I might add before merging is a visual test just to be certain that the behavior matches what we expect.

resort-diaper:
Nice work with the re-write. Extracting out the computations makes it much cleaner.

The page spec currently is not checking the right hand side of the footer which I think is where the issues are likely to show up. If it tries to put a portrait sized footer onto a landscape page the left hand side of the footer will be in the correct location but the right hand side of the footer will be in the middle of the page, as evidenced by https://github.com/asciidoctor/asciidoctor-pdf/issues/1036#issue-436099464.

mojavelinux:
I'm pretty sure there's a test which covers that scenario. If you believe otherwise, can you add a test which demonstrates the problem?

mojavelinux:
I've specifically added a test case for that use case.

PR #1071, opened 2019-05-14
Fix typo "Asciidotor"

remocrevo:


mojavelinux:
Thanks! It's amazing how those errors can hide in plain sight. Nice eye!

Issue #1087, opened 2019-06-03
Text is not printed on page header


tiagoma:
I have noticed this issue after upgrading from `asciidoctor-pdf-1.5.0.alpha.17` to `asciidoctor-pdf-1.5.0.alpha.18`:

* The page header text as specified in the YAML style file (for example in `header_recto_right_content`) is no longer being displayed in the page headers.
* Footer text is not affected and still being printed as expected.
* Images, both in header and footer, are displayed correctly.

See the output generated by each version:

[hello-1.5.0-alpha17.pdf](https://github.com/asciidoctor/asciidoctor-pdf/files/3245697/hello-1.5.0-alpha17.pdf)
![1 5 0-alpha17](https://user-images.githubusercontent.com/38113517/58774309-2e02c480-8597-11e9-978e-3365c47a83c5.png)

[hello-1.5.0-alpha18.pdf](https://github.com/asciidoctor/asciidoctor-pdf/files/3245698/hello-1.5.0-alpha18.pdf)
![1 5 0-alpha18](https://user-images.githubusercontent.com/38113517/58774316-33600f00-8597-11e9-879e-22016e908364.png)

Below the source:
```
= Hello
Hello
Hello

== Hello
Hello

<<<

== Hello
Hello
```

and the YAML style file:
```
page:
  layout: portrait
  margin: [0.75in, 1in, 0.75in, 1in]
  size: Letter
base:
  font_color: #333333
  font_family: Times-Roman
  font_size: 12
  line_height_length: 17
  line_height: $base_line_height_length / $base_font_size
vertical_spacing: $base_line_height_length
heading:
  font_color: #262626
  font_size: 17
  font_style: bold
  line_height: 1.2
  margin_bottom: $vertical_spacing
link:
  font_color: #002FA7
outline_list:
  indent: $base_font_size * 1.5
footer:
  height: $base_line_height_length * 2.5
  line_height: 1
  recto:
    center:
      content: image:octocat.jpg[pdfwidth=10mm]
    right:
      content: 'Hello {page-number}'
  verso:
    center:
      content: $footer_recto_center_content
    left:
      content: $footer_recto_right_content
header:
  height: $base_line_height_length * 2.5
  line_height: 1
  recto:
    center:
      content: image:octocat.jpg[pdfwidth=10mm]
    right:
      content: '{page-number} Hello'
  verso:
    center:
      content: $header_recto_center_content
    left:
      content: $header_recto_right_content
```

[...]

PR #1092, opened 2019-06-06
resolves #1091 apply page layout specified on page break even when break falls at page boundary

### LINKED ISSUE #1091: Page layout (landscape/portrait) stuck in landscape.
When using Asciidoctor PDF backend, 
```
[.landscape]
<<<
``` 
works to flip the page, but 
```
[.portrait]
<<<
``` 
doesn't flip it back.

The `[.landscape]` and `[.portrait]` directives are in a nested document; observed behavior is (obviously) in the document that includes it.

This is my setup (Windows 10):
```bash
$ ruby -v
ruby 2.5.5p157 (2019-03-15 revision 67260) [x64-mingw32]

$ asciidoctor -v
Asciidoctor 2.0.9 [https://asciidoctor.org]
Runtime Environment (ruby 2.5.5p157 (2019-03-15 revision 67260) [x64-mingw32]) (lc:IBM437 fs:Windows-1252 in:IBM437 ex:IBM437)

$ asciidoctor-pdf -v
Asciidoctor PDF 1.5.0.alpha.18 using Asciidoctor 2.0.9 [https://asciidoctor.org]
Runtime Environment (ruby 2.5.5p157 (2019-03-15 revision 67260) [x64-mingw32]) (lc:IBM437 fs:Windows-1252 in:IBM437 ex:IBM437)
```

mojavelinux:


dalemartin:
Layout is no longer stuck in landscape, works like a charm. Couple new issues though:
* I'm losing my running header completely
* Footer formatting is not right (I realize that's being handled elsewhere)
* Not sure but it seems like `pdfwidth=100%` doesn't work to get an image to fill the landscape width? Shouldn't it scale up to 100% by default without me specifying that anyway? In either case, it's not filling the width of the page. Maybe it's height constrained by the invisible header...

dalemartin:
Actually I'm pretty sure that last bullet is because of the height of my image. The footer (and presumably the invisible header) margins are much larger than on the portrait pages.

As a side note, is there any way of using `[%rotate]` for tables/images instead of changing the page layout?

mojavelinux:
> I'm losing my running header completely

This is a known issue in alpha.18 that will be fixed in beta.1. See #1087.

> Footer formatting is not right

I'm assuming you mean the margins are off.

> Actually I'm pretty sure that last bullet is because of the height of my image filling the vertical space.

Yes, that's correct. It still has to fit height-wise.

mojavelinux:
> As a side note, is there any way of using [%rotate] for tables/images instead of changing the page layout?

Prawn table does not support rotation.

mojavelinux:
It sounds like #1027 is crucial to fix to make page layout changes viable. I'll have a closer look at that PR.

dalemartin:
> > Footer formatting is not right
> 
> I'm assuming you mean the margins are off.
The margins are off, yes, but specifically the horizontal line and page number are short, the same width as the portrait pages. A separate issue was opened about it already #1036.

mojavelinux:
Yes, that's what I meant. :+1:

PR #1110, opened 2019-06-25
resolves #1109 exclude border width from running content area

### LINKED ISSUE #1109: Exclude border width from writable area of running content
The width of the border on the running content should be excluded when computing the writable area. Otherwise, the text or image can overlap the border.

mojavelinux:


mojavelinux:
Incredibly, this change didn't break any existing tests. The default border width must be so small that it doesn't throw off any of the visual tests. The test case uses an exaggerated border to validate the computation.

mojavelinux:
This is missing the same test for the footer. I don't believe the calculation is correct yet.

mojavelinux:
I fixed the calculation for the footer and updated the test to check the footer too.

PR #1140, opened 2019-07-08
resolves #306 allow theme to control background and border of inline code (literal)

### LINKED ISSUE #306: Add background color key to inline monospace
Currently it does not seem to be possible to create a custom theme in which a background colour can be added to inline monospace (`literal` in the theme YAML).

This is the default styling in the HTML output in the inline `code` style and it would be nice to be able to replicate this.

Default HTML CSS:

``` CSS
:not(pre) > code {
    font-size: 0.9375em;
    font-style: normal !important;
    letter-spacing: 0px;
    padding: 0.1em 0.5ex;
    word-spacing: -0.15em;
    background-color: #F7F7F8;
    border-radius: 4px;
    line-height: 1.45;
    text-rendering: optimizespeed;
}
```

As far as I can see this is currently possible in source block formatting (`code` YAML section), but not in the inline case surrounded by `.

mojavelinux:


billybooth:
Not a Ruby guy, but I believe that https://github.com/asciidoctor/asciidoctor-pdf/pull/1140/commits/8d2daab55bb391ee584df6c3f0421631bef16bc8#diff-6891f4ede6eb11fb190754a078ee93bcR134 breaks compatibility with Ruby 2.3.  If I'm not mistaken, array [union](https://ruby-doc.org/core-2.6/Array.html#method-i-union) was added in 2.6.  I only noticed because we were running into it with the [docker container](https://hub.docker.com/r/asciidoctor/docker-asciidoctor/) which uses Ruby 2.5.5p157.  In our case, it was easy enough to tweak our .adoc to work around the issue, but just thought I should give you a heads up.

Let me know if you'd like me to report this as an issue.

mojavelinux:
Technically Ruby 2.3 is now EOL. But we should still be supporting 2.4.

But what's confusing is that we do test on 2.3. See https://travis-ci.org/asciidoctor/asciidoctor-pdf. Are we missing a test?

PR #1166, opened 2019-07-19
resolves #1164 add support for default theme alignment for tables

### LINKED ISSUE #1164: Default table alignment from theme
Currently we have
```
image:
  align: left | center | right
caption:
  align: left | center | right
```
Proposing a matching key for tables:
```
table:
  align: left | center | right
```
Should be able to override using the current `align="left | center | right"` property for individual placement.

habamax:
fix for #1164


mojavelinux:
This looks good to me! Great job.

Issue #368, opened 2015-12-09
Add support for inline roles to theme


mojavelinux:
Allow the theme to control the styling for roles on inline elements. When a role is detected on an inline element, it consults the theme to get styling information to merge into the default. This effectively brings styling using CSS classes to inline elements.

Example:

```
[.type]#name of type#
```

mojavelinux:
We also may want to define some built-in, semantic role names as discussed in https://github.com/asciidoctor/asciidoctor/issues/1664.

mattwynne:
Given that this issue is still open, is there any way at the moment for me to style individual pages or other elements in a PDF, e.g. adding a different background image to each level 1 section?

No criticism or pressure meant - I work a lot on OSS myself and I know you need to be free to just get around to things when you can. I'm just new to asciidoctor and trying to get my head around the scope of what's possible. In general it seems like an awesome piece of work so far, so thanks very much.

mattwynne:
@mojavelinux I dunno how hard this is or how much bandwidth you have, but I am feeling the pain of this right now and would be motivated to have a crack at implementing it myself if you were able to give me some pointers.

I'm going to experiment with using a custom converter (as mentioned in #337) for now.

mojavelinux:
I think the open pull request for this should at least get you part of the way there. See https://github.com/asciidoctor/asciidoctor-pdf/pull/350

What we need is someone to take that to completion. It will solve a lot of open requests.

mojavelinux:
The tricky part is deciding how to configure the properties (i.e., keys) of the role in the theme so they don't collide with built-in keys.

I think the simplest way to do this is to reserve `role` as a top-level category. The custom roles would then be defined inside that category.

The first subkey level represents the role name (no hyphens or underscores allowed), and the keys under those roles are the concrete theming properties such as `font-color`.

Here's an example of some roles for text colors:

```yaml
role:
  red:
    font-color: ff0000
  green:
    font-color: 00ff00
  blue:
    font-color: 0000ff
```

PR #1177, opened 2019-07-28
resolves #459 add support for big and small roles on phrases

### LINKED ISSUE #459: add support for big and small roles on phrases
For example:

```
[small]#some small text#
[big]#some big text#
```

in PDF appears as a regular sized text.

Adding

```
base:
  font_size_large: $base_font_size * 1.5
  font_size_small: $base_font_size * 0.75
```

to theme description did not help

mojavelinux:


mojavelinux:
Now that there's an implementation for #368, switch to using built-in role names instead of hard-coded behavior.

PR #1201, opened 2019-08-07
Add explanation for PDF import using image::[]

nuket:
Took much longer to figure this out than I wanted, maybe a small write-up will help others.


mojavelinux:
Thanks!

FYI, you don't need the manual page breaks. Importing a PDF always creates a dedicated page.

Could you reformat using sentence-per-line (ventilated prose)?

nuket:
Done.

mojavelinux:
Thanks @nuket!

PR #1239, opened 2019-08-30
resolves #1223 allow to define text background and border on inline marks

### LINKED ISSUE #1223: text background color for custom inline roles
Is it possible to change text background color in asciidoctor-pdf?

Mogztter:
resolves #1223

### DIFF
--- a/lib/asciidoctor-pdf/formatted_text/transform.rb
+++ b/lib/asciidoctor-pdf/formatted_text/transform.rb
@@ -20,6 +20,11 @@ class Transform
     'font_family' => :font,
     'font_size' => :size,
     'font_style' => :styles,
+    'background_color' => :background_color,
+    'border_offset' => :border_offset,
+    'border_radius' => :border_radius,
+    'border_width' => :border_width,
+    'border_color' => :border_color,
   }
   #DummyText = ?\u0000
 
@@ -321,6 +326,7 @@ class Transform
         styles << :strikethrough
       else
         fragment.update(@theme_settings[class_name]) {|k, oval, nval| k == :styles ? oval.merge(nval) : oval } if @theme_settings.key? class_name
+        (fragment[:callback] ||= []) << TextBackgroundAndBorderRenderer
       end
     end if attrs.key?(:class)
     fragment.delete(:styles) if styles.empty?
--- a/lib/asciidoctor-pdf/formatted_text/transform.rb
+++ b/lib/asciidoctor-pdf/formatted_text/transform.rb
@@ -326,7 +326,9 @@ class Transform
         styles << :strikethrough
       else
         fragment.update(@theme_settings[class_name]) {|k, oval, nval| k == :styles ? oval.merge(nval) : oval } if @theme_settings.key? class_name
-        (fragment[:callback] ||= []) << TextBackgroundAndBorderRenderer
+        if fragment[:background_color] || (fragment[:border_color] && fragment[:border_width])
+          ((fragment[:callback] ||= []) << TextBackgroundAndBorderRenderer).uniq!
+        end 
       end
     end if attrs.key?(:class)
     fragment.delete(:styles) if styles.empty?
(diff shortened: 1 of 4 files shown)


lib/asciidoctor-pdf/formatted_text/transform.rb
```diff
@@ -321,6 +326,7 @@ def build_fragment(fragment, tag_name, attrs = {})
         styles << :strikethrough
       else
         fragment.update(@theme_settings[class_name]) {|k, oval, nval| k == :styles ? oval.merge(nval) : oval } if @theme_settings.key? class_name
+        (fragment[:callback] ||= []) << TextBackgroundAndBorderRenderer
```
mojavelinux: We don't always want to add this callback, as it isn't always necessary. It should only be added when either a background color or both a border color and border width are set.

In other words:

```
background_color || (border_width && border_color)
```

So we'll need some key detection.

Perhaps something like:

```rb
if @theme_settings.key? class_name
  fragment.update(@theme_settings[class_name]) {|k, oval, nval| k == :styles ? oval.merge(nval) : oval }
  if fragment[:background_color] || (fragment[:border_color] && fragment[:border_width])
    ((fragment[:callback] ||= []) << TextBackgroundAndBorderRenderer).uniq!
  end 
end
```

spec/text_formatter_spec.rb
```diff
@@ -238,5 +238,20 @@
       (expect text[1][:font_size]).to eql 12
       (expect text[2][:font_size].to_f.round 2).to eql 10.0
     end
+
+    it 'should allow theme to override background and border for custom role' do
```
mojavelinux: Add `integration: true` to the test description.

lib/asciidoctor-pdf/formatted_text/transform.rb
mojavelinux: This should have been marked request changed. I clicked the wrong button.

lib/asciidoctor-pdf/formatted_text/transform.rb
```diff
@@ -20,6 +20,11 @@ class Transform
     'font_family' => :font,
     'font_size' => :size,
     'font_style' => :styles,
+    'background_color' => :background_color,
+    'border_offset' => :border_offset,
+    'border_radius' => :border_radius,
+    'border_width' => :border_width,
+    'border_color' => :border_color,
```
mojavelinux: Let's sort the border property names.

lib/asciidoctor-pdf/formatted_text/transform.rb
Mogztter: I initially did that but I realized that `TextBackgroundAndBorderRenderer` is already checking if the background and/or the border are defined.
Sure it's not really efficient since the callback will be called and will do nothing.

I will add the condition 👍

Mogztter: > both a border color and border width are set.

There's no default color for the border? If I just define a border width and radius it will do nothing?

spec/text_formatter_spec.rb
Mogztter: done

lib/asciidoctor-pdf/formatted_text/transform.rb
Mogztter: Ok, done.

Mogztter:
> Nice job on the visual test!

Actually it's really pleasant to write tests, the `visually_match` and `to_pdf_file ` are very useful.

lib/asciidoctor-pdf/formatted_text/transform.rb
mojavelinux: In this context, that's correct.

There is a default border color in other places, but I don't think it makes sense to apply it here.

mojavelinux: > Sure it's not really efficient since the callback will be called and will do nothing.

While our renderer is inefficient, it's Prawn that isn't. When you register callbacks, it does all sorts of extra calls (as I recall).

mojavelinux:
Fantastic! Thanks!

PR #1256, opened 2019-09-05
resolves #395 document how to extend the converter

### LINKED ISSUE #395: Document how to extend the converter
The nature of hackable typesetting is that the converter is opinionated and thus may not produce exactly the output that someone wants. The theme support is there to cover basic customizations (fonts, colors, spacings, etc.), but it can only ever go so far (short of recreating CSS in Prawn). At some point, it will be necessary to extend the converter in order to meet layout requirements. We should document how this task can be accomplished either in the theming guide or a dedicated guide on the subject.

mojavelinux:


mojavelinux:
@habamax How do you feel about this doc addition? Does it make sense?

@Mogztter Your thoughts? Anything missing?

docs/theming-guide.adoc
```diff
@@ -4363,6 +4364,126 @@ For more information about source highlighting with Rouge, refer to the http://r
...
+  # overrides go here
+end
+----
+
+As it stands, the converter doesn't do anything differently than the primary converter because we haven't yet overridden any of its methods.
+Let's do that now.
+
+The methods that handle conversion follow the pattern `convert_<name>` for block elements (e.g., `convert_example`) and `convert_inline_<name>` for inline elements (e.g., `convert_inline_anchor`), where `<name>` is the name of the element.
+
+When you overide a block element, you write PDF objects directly to the Prawn Document (the current context).
```
Mogztter: you override
```suggestion
When you override a block element, you write PDF objects directly to the Prawn Document (the current context).
```

docs/theming-guide.adoc
```diff
@@ -4363,6 +4364,126 @@ For more information about source highlighting with Rouge, refer to the http://r
...
+    typeset_text title, (calc_line_metrics 1.5), color: 'FFFFFF', inline_format: true, align: :center, size: 42
+    page.instance_variable_set :@imported_page, true
+  end
+----
+
+The method `typeset_text` and `calc_line_metrics` are provided by Asciidoctor PDF to make writing text easier.
+If you wanted, you could just use the low-level `text` method provided by Prawn.
+
+To find all the available methods to override, consult the https://www.rubydoc.info/github/asciidoctor/asciidoctor-pdf/Asciidoctor/Pdf/Converter[API docs].
+For deeper examples of how to override the behavior of the converter, refer to the extended converter in the https://github.com/mraible/infoq-mini-book/blob/master/src/main/ruby/asciidoctor-pdf-extensions.rb[InfoQ Mini-Book template].
```
Mogztter: We should probably use a SHA1 to make sure that the file will be resolved.

docs/theming-guide.adoc
```diff
@@ -4363,6 +4364,126 @@ For more information about source highlighting with Rouge, refer to the http://r
...
+Let's say we want to override the kbd element.
+
+[source,ruby]
+----
+  def convert_inline_kbd node
+    %(<strong><color rgb="AA0000">#{(node.attr 'keys').join ' + '}</color></strong>)
+  end
+----
+
+Refer to the primary converter to discover the pseudo-HTML you can use for inline elements.
```
Mogztter: Is there an official documentation that explain which HTML elements are supported?

Mogztter:
I've added a few comments but overall it's very well written.
I can't think of anything missing 👍

docs/theming-guide.adoc
```diff
@@ -4363,6 +4364,126 @@ For more information about source highlighting with Rouge, refer to the http://r
...
+The theme support is there to provide basic customizations (fonts, colors, borders, spacing, etc.).
+But it can only go so far.
+At some point, it becomes necessary to extend the converter to meet advanced design requirements.
+
+Before you dive into extending this converter, you'll need to understand how to use Prawn.
+The article https://www.sitepoint.com/hackable-pdf-typesetting-in-ruby-with-prawn/[Hackable PDF Typesetting in Ruby with Prawn] gives a crash course in how to create your first PDF document containing text, graphics, and images with Prawn.
+That article is essential reading for working with Asciidoctor PDF, which uses many of the same operations (as well as many helpful add-ons).
+Once you feel comfortable with Prawn, you're ready to extend the converter.
+
+=== Defining the Extended the Converter
```
habamax: Not sure if I get this. 
But I am not native though.

habamax: Probably should be "Defining the Extended Converter" ?

habamax:
> @habamax How do you feel about this doc addition? Does it make sense?
> 
It is well written and might be really helpful if I want to do converter customization. On the other hand I feel this addition should be part of another document -- Customization guide or Plugin guide? Like the target audience of Theming (just settings in form of yaml) is a lot wider than what is written here.

docs/theming-guide.adoc
mojavelinux: Correct.

docs/theming-guide.adoc
mojavelinux: Since that project has mostly stabilized, I expect this link to be good for years.

docs/theming-guide.adoc
mojavelinux: :+1:

docs/theming-guide.adoc
mojavelinux: I'll go ahead and add this information.

mojavelinux:
> On the other hand I feel this addition should be part of another document -- Customization guide or Plugin guide?

I agree, though we're holding off adding any additional pages until the documentation has been migrated to Antora (which is in process). Then it will become it's own page (as will a lot of this document).

mojavelinux:
Thanks for reviewing!

PR #1261, opened 2019-09-06
resolves #1204 use consistent line height for list items and toc entries if text is entirely monospace

### LINKED ISSUE #1204: Unordered list bullet + literal = wrong vertical bullet alignment
tested in beta2 and current master

```
= test

== test 1
* hello
** world
** `world`
** world `world`
** w``orld``
* hello
* `hello`
* hello
```

With everything default outputs pdf:

image

Note if list item starts with inline literal the bullet is misaligned. 

PS, boxes and arrows are added afterwards to illustrate the issue.

mojavelinux:


mojavelinux:
@habamax want to give this a try?

habamax:
Looks good for me. Tested with 3 docs that had this issue (removing handwritten `{zwsp}`s)

habamax:
Although the name of the fix "...if text is entirely monospace" is not true for me -- it also works if text is not entirely monospaced:

```
* `I` -- Inserted
* `U` -- Updated
```

mojavelinux:
Thanks for confirming!

> it also works if text is not entirely monospaced:

I'm not able to reproduce this. But I guess it doesn't really matter since the fix would solve it anyway.

PR #1262, opened 2019-09-06
adds documentation about new keys in role category

juracy:
After #1239 new keys were added and not documented: background-color, border-color, border-offset, border-radius and border-width


mojavelinux:
You're absolutely right. Thanks for updating the docs!

PR #1271, opened 2019-09-12
resolves #1219 pass styles for inline elements downwards when parsing

### LINKED ISSUE #1219: Pass text styles downward (from parent to child)
Currently, text styles (i.e., formatting) are applied on the traversal out of the hierarchy (in `Asciidoctor::PDF::FormattedText::Transform`). This means that styles are not properly inherited, and sometimes the decision about which style to apply is ambiguous. Instead, the styles should be passed downward through the hierarchy (from parent to child) so that the style on the inner-most element always wins out.

mojavelinux:
- allowing role to override default styles for element

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -7,6 +7,7 @@ For a detailed view of what has changed, refer to the {uri-repo}/commits/master[
 
 == Unreleased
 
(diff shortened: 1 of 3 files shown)


mojavelinux:
Thanks for reviewing! Let's go for it.

PR #1311, opened 2019-10-09
resolves #1310 allow theme to set font color to transparent

### LINKED ISSUE #1310: Add support for transparent font color
The transparent keyword is not recognized when used as the font color. Add support for this keyword.

This will require patching the text arranger in Prawn to set the transparency when the fragment color is transparent. This may also open the door for partial transparency in the future, since it requires overriding the same method.

mojavelinux:


mojavelinux:
This has side effects, so I've decided to forgo this enhancement for now.

PR #1321, opened 2019-10-12
resolves #1320 set default footer content in base theme

### LINKED ISSUE #1320: Specifying none as value of running content doesn't work as documented
The theming guide claims that using the keyword none will clear the running content for a specific periphery and side.

> If content is not specified for the running footer, the page number (i.e., {page-number}) is shown on the left on verso pages and the right on recto pages. You can disable this behavior by defining the attribute nofooter in the AsciiDoc document header or by defining the key `footer-<side>-content: none` in the theme.

For example:

```
footer:
  recto:
    right:
      content: none
```

In practice, this results in the string "none" being used.

The behavior either needs to be fixed or the documentation updated.

mojavelinux:
- remove logic to process `footer_<side>_content: none` key

### DIFF
--- a/docs/theming-guide.adoc
+++ b/docs/theming-guide.adoc
@@ -4226,9 +4226,6 @@ IMPORTANT: If you don't specify a height for either the header or footer key, it
 
 If you define running header and footer content in your theme (including the height), you can still disable this content per document by setting the `noheader` and `nofooter` attributes in the AsciiDoc document header, respectively.
 
-If content is not specified for the running footer, the page number (i.e., `\{page-number}`) is shown on the left on verso pages and the right on recto pages.
-You can disable this behavior by defining the attribute `nofooter` in the AsciiDoc document header or by defining the key `footer-<side>-content: none` in the theme.
-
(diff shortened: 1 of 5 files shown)


docs/theming-guide.adoc
```diff
@@ -4226,9 +4226,6 @@ IMPORTANT: If you don't specify a height for either the header or footer key, it
 
 If you define running header and footer content in your theme (including the height), you can still disable this content per document by setting the `noheader` and `nofooter` attributes in the AsciiDoc document header, respectively.
 
-If content is not specified for the running footer, the page number (i.e., `\{page-number}`) is shown on the left on verso pages and the right on recto pages.
```
lurch: Even though it's no longer explicitly added in `converter.rb`, I guess any theme extending base or default will still see this "default page numbers" behaviour?
Perhaps it's still worth documenting e.g. `footer_verso_left_content: null` ? :man_shrugging:

mojavelinux: Sure. I've gone ahead and updated the docs with this information.

PR #1339, opened 2019-10-21
Bump treetop to 1.6

snecklifter:
Treetop 1.5 is now five years old and some distributions have been
running with treetop 1.6 without issue for some time.

### DIFF
--- a/asciidoctor-pdf.gemspec
+++ b/asciidoctor-pdf.gemspec
@@ -45,8 +45,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'safe_yaml', '~> 1.0.0'
   s.add_runtime_dependency 'thread_safe', '~> 0.3.0'
   s.add_runtime_dependency 'concurrent-ruby', '~> 1.1.0'
-  # For our usage, treetop 1.6 is slower than treetop 1.5
-  s.add_runtime_dependency 'treetop', '~> 1.5.0'
+  s.add_runtime_dependency 'treetop', '~> 1.6.0'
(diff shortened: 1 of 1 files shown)


snecklifter:
@mojavelinux would you mind considering this PR please? I have no data to back up the performance hit but wondering if this can be re-visited.

mojavelinux:
We've held back the version of treetop because treetop 1.6 is slower. I don't know why it's slower, but it's substantially slower. So if we can answer why it is slower, and whether it will ever be faster than 1.5 was, then I'm ready to proceed. That's the hump we need to get over.

mojavelinux:
I've tested it again and it's still slower. Here's the difference in a test parse of 50 runs.

```
1.5.3 - 0.001674898
1.6.10 - 0.002010059
```

That's an increase of 20%, which is huge. Given that speed is already an issue when creating PDFs, I'm very hesitant to add overhead for no good reason.

mojavelinux:
It does appear that the parsing is such a small part of what Asciidoctor PDF does that it doesn't impact the speed of the test suite. So while this increase is huge, it's impact doesn't seem to be noticeable.

What I don't understand is why treetop got so much slower. It still generates the same parse code. So where's the overhead coming from? Are they doing anything to address it?

slonopotamus:
> it's impact doesn't seem to be noticeable

So, is there something that prevents this from being merged then?

mojavelinux:
Frankly, I just don't care anymore about the line I was trying to hold on this, so let's just upgrade.

Issue #1206, opened 2019-08-12
Use specialized line break strategy for Chinese and Japanese


mojavelinux:
Line break rules for Latin-based languages such as English and French are also being applied to Chinese and Japanese. However, Chinese and Japanese don't use spaces (at least not in the same way). While Latin-based languages have spaces between words where line breaks can occur, Chinese and Japanese are written without spaces in which a line break can occur between any two characters. Chinese and Japanese also use different punctuation for pause, full stop, and dash. These need to be taken into account.

While this isn't so much of a problem when the text is written exclusively in a CJK language (since the line break will be forced once the line is full), it becomes a problem when the text is mixed with another language such as English. All of a sudden, huge gaps appear because the groups of CJK languages get treated as a single "word".

Here's an example:

```
AsciiDoc 是一个人类可读的文件格式，语义上等同于 DocBook 的 XML，但使用纯文本标记了约定。可以使用任何文本编辑器创建文件把 AsciiDoc 和阅读“原样”，或呈现为HTML 或由 DocBook 的工具链支持的任何其他格式，如 PDF，TeX 的，Unix 的手册页，电子书，幻灯片演示等。
```

When rendered with Asciidoctor PDF, huge gaps appear in the line. This can be partially mitigated by changing the text alignment from left to justify, but then the gaps are just shifted to the end of the line.

The correct fix is to allow a line break between any two CJK characters, as long as one of the characters is not punctuation.

To activate this specialized logic, the author must set the `scripts` attribute in the document header to `cjk`.

Related issue: #82.

mojavelinux:
At first I though it would be possible to patch the method in Prawn that scans for line break opportunities. However, that logic is very difficult to override (not to mention understand). A simpler approach, which @chloerei proposed, is to modify the string being typeset by inserting zero-width spaces at line break opportunities. While that may not be the most elegant solution, it gets us a solution that we can use today and leaves room for better solutions to come along...including a fix in Prawn itself.

Here's the crux of that logic:

```
string = string.gsub %r/(?=[\u3000\u30a0-\u30ff\u3040-\u309f\p{Han}\uff00-\uffef])/, ZeroWidthSpace
```

* `\u3000` is the ideograph space character
* `\u30a0-\u30ff` is Hirgana (\p{Hiragana} is incomplete)
* `\u3040-\u309f` is Katakana
* `\p{Han}` are the unified CJK ideographs
[...]

PR #1355, opened 2019-10-30
resolves #1359 break CJK characters in table when scripts attribute is `cjk`

### LINKED ISSUE #1359: Apply CJK line breaks to contents of table cell
When the `scripts` attribute is set to `cjk`, the CJK line break rules are not applied to the contents of a table cell.

Gasol:
Fix https://github.com/chloerei/asciidoctor-pdf-cjk/issues/4, Also see https://github.com/asciidoctor/asciidoctor-pdf/issues/1206 for details

Require fileutils explictly to fix following errors when run command with `rake spec`

    An error occurred in a `before(:suite)` hook.
    Failure/Error: FileUtils.mkdir_p output_dir

    NameError:
      uninitialized constant FileUtils
      Did you mean?  FileTest
    # ./spec/spec_helper.rb:187:in `block (2 levels) in <top (required)>'

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -8,6 +8,7 @@ For a detailed view of what has changed, refer to the {uri-repo}/commits/master[
 == Unreleased
 
 * allow theme to override styles of caption on admonition blocks (#561)
+* if value of scripts attribute is cjk, break lines between any two CJK characters except punctuation in table cells (#1359) (*gasol*)
 
 == 1.5.0.beta.7 (2019-10-29) - @mojavelinux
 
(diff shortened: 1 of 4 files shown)


mojavelinux:
Thanks for pointing this out and for submitting a fix! I created an issue so we can reference this change in the CHANGELOG.

mojavelinux:
> Require fileutils explictly to fix following errors when run command with rake spec

This should not be necessary. Prawn already includes fileutils. You should be executing tests using `bundle exec rake spec`.

Gasol:
> bundle exec rake spec

I'm ok, But I think require package explicitly is better than implicit.

mojavelinux:
Since we only use it in the test suite anyway, I'll autoload it. But I want to do it in a separate change.

Issue #1326, opened 2019-10-15
Allow custom (inline) role to control text decoration property


mojavelinux:
Currently the text-decoration property on a custom (inline) role is not honored. Support should be added without interfering with the font-style.

PR #1363, opened 2019-11-01
adds documentation about text-decoration in role category

juracy:
After #1326 text-decoration was added and not documented


mojavelinux:
Thanks!

PR #1404, opened 2019-11-24
honor background-color style on span to support highlighter styles

mojavelinux:


mojavelinux:
merged as 21114af299181d066c2502c70e9963954767bc95

PR #1408, opened 2019-11-24
resolves #1407 add a keep-together role on open blocks

### LINKED ISSUE #1407: Allow to define a keep together open block to avoid page break between related content
The idea is to be able to create a "keep-together" block to avoid page break:

```
[.keep-together]
--
The following image shows the stellar evolution of a high-mass star:

image::star[]
--
```

In the above document, we want to avoid a page break between the introductory paragraph and the image.
Similarly, we could use this block with other types of content:

```
[.keep-together]
--
To install Antora, open a terminal and type:

 $ npm i -g @antora/cli@2.2 @antora/site-generator-default@2.2

IMPORTANT: The `@` at the beginning of the package name is important.
It tells `npm` that the `cli` package is located in the `antora` group.
If you omit this character, `npm` will assume the package name is the name of a git repository on GitHub.
The second `@` offsets the requested version number.
Only the major and minor segments are specified to ensure you receive the latest patch update.
--
```

Mogztter:


lib/asciidoctor/pdf/converter.rb
```diff
@@ -933,9 +933,17 @@ def convert_open node
       # TODO process abstract child even when partintro has multiple blocks
       convert_abstract node.blocks[0]
     else
-      add_dest_for_block node if node.id
-      layout_caption node.title if node.title?
-      convert_content_for_block node
+      if node.has_role? 'keep-together'
```
mojavelinux: To be consistent with AsciiDoc.py and other converters, this should instead check for the unbreakable role.

```rb
if node.option? 'unbreakable'
```

lib/asciidoctor/pdf/converter.rb
```diff
@@ -933,9 +933,17 @@ def convert_open node
...
+        keep_together do
```
mojavelinux: We can make this DRY by using `keep_together_if`, like:

```rb
keep_together_if node.option? 'unbreakable' do
```

To avoid footnotes from being duplicated, it's also necessary to add this line at the start of the block:

```rb
push_scratch doc if scratch?
```

and this line to the end:

```rb
pop_scratch doc if scratch?
```

In all, that looks like:

```rb
doc = node.document
keep_together_if node.option? 'unbreakable' do
  push_scratch doc if scratch?
  add_dest_for_block node if node.id
  layout_caption node.title if node.title?
  convert_content_for_block node
  pop_scratch doc if scratch?
end
```

lib/asciidoctor/pdf/converter.rb
Mogztter: > this should instead check for the unbreakable role.

You mean option?

> You can use the unbreakable option to instruct the browser not to break a block element. The following image and it’s caption will be kept together the printed page:
> ```
> [options="unbreakable"]
> .Tiger block image
> image::images/tiger.png[Tiger image]
> ```
https://www.methods.co.nz/asciidoc/faq.html#_how_can_i_control_page_breaks_when_printing_html_outputs

lib/asciidoctor/pdf/converter.rb
Mogztter: Thanks for your review.
I wasn't aware of this issue with footnotes. I will add footnotes in my test to make sure that they are not duplicated.

lib/asciidoctor/pdf/converter.rb
mojavelinux: Yep! Or the shorthand for it:

```
[%unbreakable]
```

lib/asciidoctor/pdf/converter.rb
mojavelinux: It's a bit weird, and I wish it was handled by keep_together. But that would require some refactoring. For now, we just add it where keep_together is used.

spec/example_spec.rb
```diff
@@ -1,6 +1,30 @@
 require_relative 'spec_helper'
 
 describe 'Asciidoctor::PDF::Converter - Example' do
+  it 'should keep block together when block has the unbreakable option' do
+    to_file = to_pdf_file <<~EOS, 'example-unbreakable-option.pdf'
+    Make it rain.footnote:[money]
```
Mogztter: Even without `push_scratch doc if scratch?` and `pop_scratch doc if scratch?` the footnotes are not duplicated so I'm not sure how to test this case.

mojavelinux: The footnotes have to be inside the open block.

Mogztter: I did add a footnote before the open block, inside the open block and after the open block :thinking:

mojavelinux:
I think the block also has to break across pages in the scratch document,
at which point it will try to render it a second time.

Mogztter:
Oh I see, that's a good thing to test (ie. when an unbreakable block does not fit on a page). I will add a test.

spec/example_spec.rb
mojavelinux: Aha. The scenario is that the footnote is in the open block, but not inside another nested block. In the test, the footnote is in an admonition block, which masks the problem.

PR #1409, opened 2019-11-25
resolves #1405 ignore case when sorting terms in index

### LINKED ISSUE #1405: Sorting of index terms should be case insenstivie
When entries within a category in the index are sorted, the sort should be case insensitive. This is an accepted practice in the publishing industry.

For example:

```
A
anchor
AsciiDoc
Asciidoctor
authoring
```

mojavelinux:


mojavelinux:
merged as b7a19e564e7ca5b47babadd558085711e5b036e7

PR #1495, opened 2020-01-19
resolves #1287 continue border, background, and column rule of admonition block on subsequent pages

### LINKED ISSUE #1287: Handle case when admonition block spans more than one page
When the height of an admonition block exceeds one page, the line between the label and the content extends into the footer on the first page, then stops. The line should break with the content and continue onto subsequent pages.

mojavelinux:


mojavelinux:
merged as 2527c876c5a0ca7d62796c98ff2efd553d255c60

PR #1496, opened 2020-01-20
Add GitHub Actions script

slonopotamus:
See workflow runs here: https://github.com/slonopotamus/asciidoctor-pdf/actions

This PR cannot be taken as-is because it exposes several problems:

1. [jaro_winkler (rubocop dependency) crashes compiler on Ruby 2.3 + Windows](https://github.com/slonopotamus/asciidoctor-pdf/runs/398410610#step:4:163)
2. [Several tests fail to pass on Windows](https://github.com/slonopotamus/asciidoctor-pdf/runs/398467720#step:4:121). Mostly because they're trying to execute files that are not executable from Windows point of view, but there are some failures that need additional investigation.
WRT running non-executable things, I suggest using `bundle exec <gem_binary>` instead of just `<gem_binary>`, this is more cross-platform.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,23 @@
+name: CI
+on: [push, pull_request]
+jobs:
+  build:
+    strategy:
+      matrix:
+        ruby: [jruby, 2.4, 2.7]
+    runs-on: windows-latest
+    env:
+      PYGMENTS_VERSION: ~> 1.2.0
+    steps:
+      - name: Set up Ruby
+        uses: eregon/use-ruby-action@v1.10.0
+        with:
+          ruby-version: ${{ matrix.ruby }}
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Build
+        run: |
+          gem install bundler
+          bundle install --jobs 4 --retry 3
+      - name: Test
+        run: bundle exec rake spec


mojavelinux:
> Mostly because they're trying to execute files that are not executable from Windows point of view, but there are some failures that need additional investigation.

I wonder why these pass on AppVeyor.

Currently, AppVeyor only runs 2.7. At the very least, I should add 2.3 until we make the switch over.

mojavelinux:
It turns out, the AppVeyor builds aren't even running. ooops!

mojavelinux:
> WRT running non-executable things, I suggest using bundle exec <gem_binary> instead of just <gem_binary>, this is more cross-platform.

Yes and no. The problem with running the CLI tests that way is that it's not a pure test. It can produce false positives as the result of being tainted by the Bundler environment. This situation is already handled in the test suite. Some tests require the Bundler environment, others don't. By default, the Bundler environment is inherited (via environment variables) when executing a command. But the run_command helper has a way to disable this to run certain tests outside of Bundler.

slonopotamus:
Whoops)

mojavelinux:
I'm not ready to trust these results on Windows. Those errors, esp the CLI ones, seem odd to me. I'll need to investigate.

mojavelinux:
Hmm, looks like now that AppVeyor is running, we have similar issues.

slonopotamus:
My point is that when [you call Gem.bin_path](https://github.com/asciidoctor/asciidoctor-pdf/blob/master/spec/spec_helper.rb#L284-L286), it will give you a path to ruby file with a shebang. You cannot simply spawn a process out of that because on Windows, there is no machinery to handle shebangs.

When you're installing via bundler/gem, it creates batch wrappers, that's why you can run `asciidoctor` from command line:

image

For the reference, contents of these wrappers (they are all identical):
```bat
@ECHO OFF
@"%~dp0ruby.exe" "%~dpn0" %*
```

This "cheat" only works because of the way how `cmd.exe` decides what to run. It actually runs `asciidoctor.bat` when you type `asciidoctor`.

But when you're spawning processes programmatically, there's nobody in-between you and OS to use the wrapper.

mojavelinux:
> My point is that when you call Gem.bin_path , it will give you a path to ruby file with a shebang. You cannot simply spawn a process out of that because on Windows, there is no machinery to handle shebangs.

Yep, I gotcha. But I had already solved this in the Asciidoctor core test suite. I just forgot about it. You need to prepend `ruby` (the full path) to the command.

mojavelinux:
> But when you're spawning processes programmatically, there's nobody in-between you and OS to use the wrapper.

I had forgotten about that. Fortunately, my past self had already figured out a solution ;)

slonopotamus:
Okay, now it passes.

Notes:
1. Unable to test 2.3 on Windows. Compiler crashes when compiling jaro_winkler (rubocop dependency) o_O
2. Didn't add asciidoctor-1.5.3 to the matrix.

mojavelinux:
> Unable to test 2.3 on Windows. Compiler crashes when compiling jaro_winkler (rubocop dependency)

Bizarre that this doesn't fail on AppVeyor. Then again, there are different ways to install Ruby on Windows, so perhaps the installations are just different. Though the one on AppVeyor is the 32-bit version. Maybe that matters.

.gitattributes
```diff
@@ -0,0 +1 @@
+*.pdf -text
```
mojavelinux: This shouldn't be necessary. A pdf document isn't a binary file. It's a text file with encoded strings in it.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Set up Ruby
+        uses: eregon/use-ruby-action@master
+        with:
+          ruby-version: ${{ matrix.ruby }}
+      - name: Build
+        run: |
+          gem install bundler
+          bundle install --jobs 4 --retry 3
+      - name: Test
+        run: bundle exec ruby -w $(bundle exec ruby -e "print Gem.bin_path 'rake', 'rake'") spec
```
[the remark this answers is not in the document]
mojavelinux: (I wish rake honored the -w flag).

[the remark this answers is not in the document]
mojavelinux: Normally, setting RUBYOPT=-w would work, except it breaks the CLI tests. That's why I don't use it.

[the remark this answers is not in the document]
slonopotamus: I'd say these are of equal complexity, but as you wish.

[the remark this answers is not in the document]
mojavelinux: It's not just about complexity. It's the right location from which to load the bin script because it's the real bin script after gem install. The one inside the gem is not guaranteed to be a full bin script (as the gem may do processing on it during packaging).

.gitattributes
mojavelinux: Yes, that seems correct.

AppVeyor must be setting this globally because it doesn't affect the builds on AppVeyor. Sure enough: https://github.com/appveyor/docker/commit/68e82c80e0506927e68fb3c22f09ea0d36bd8974

mojavelinux: I really dislike that git clients still try to change the newline character to CRLF. There's really no reason to be using that line ending at this point. It's just unnecessary lashing of the file contents, and would break the linter too I would think.

slonopotamus: > I really dislike that git clients still try to change the newline character to CRLF. There's really no reason to be using that line ending at this point.

You can add`* -text` to `.gitattributes`. It means "Git, stop touching my EOLs".

mojavelinux: I'd prefer the setting that tells git not to mess with the files. Though I could also see the risk of someone committing a file with CRLF in it. So enabling both settings might be best.

slonopotamus: They're mutually exclusive. Documentation: https://git-scm.com/docs/gitattributes#_checking_out_and_checking_in

slonopotamus: Okay, changed this to `*.pdf eol=lf`.

mojavelinux: It's not binary data. It's just an encoded stream.

mojavelinux: It's encoded in FlateDecode. I mean, it's not plain text to be sure. But it's still character data.

mojavelinux:
Now that I got AppVeyor up and running, we can compare runtimes w/ GitHub Actions. Given that we are so resource constrained on AppVeyor, having GitHub Actions handle Windows builds seems to be an easy choice. Travis is really slow at running the Linux builds, though it does allows us to install the packages we need to test the gs integration.

I'm not convinced the macOS builds are necessary. There is really 0 difference in how a gem runs on macOS vs Linux. That just seems like overhead. I could maybe see running one Ruby just to touch it, but it's not going to add much value otherwise.

.gitattributes
slonopotamus: I mean, it's sequence of bytes that have CRLF sequence in them and it would be an error to convert CRLF->LF in this sequence. `-text` is the only proper solution for this.

slonopotamus:
Dropped MacOS from matrix.

I'm not going to persuade you into using GitHub Actions. Even if the only effect of this PR is that you've fixed AppVeyor builds, this is good enough for me.

It is unfortunate that Ruby 2.3+Windows fails, this needs to be reported to jaro_winkler project.

mojavelinux:
> I'm not going to persuade you into using GitHub Actions. 

Oh, I think we should use it. I'm just trying to determine how much of it and when. I'm really not worried about the Windows Ruby 2.3 issue since 2.3 on Windows is really bad anyway. We can just test 2.4.

slonopotamus:
Discussion of [ asciidoctor-pdf Packaging should install bin script named asciidoctor-pdf test failure](https://github.com/slonopotamus/asciidoctor-pdf/runs/408612068?check_suite_focus=true#step:5:10): https://github.com/asciidoctor/asciidoctor-pdf/commit/9e1ca816f1fe73f5d90e305925f5de7a32c84440#r36962448

mojavelinux:
Aha, I see the problem:

```
bundle install --jobs 4 --retry 3
```

That needs to be:

```
bundle install --path=.bundle/gems --jobs 4 --retry 3
```

The tests currently assume that the gems are installed inside the project.

mojavelinux:
No, that's not it. The issue is that GitHub Actions doesn't use RVM. And there are subtle differences about where RVM installs things.

mojavelinux:
I revved up a Docker image and duplicated what GitHub Actions does. It's doing something non-standard with Bundler. The bin script for the current project is not getting installed in the bin directory, which is exactly what the test is trying to validate.

This is solved by adding `--path .bundle/gems` when invoking bundler. That puts it into the right mode.

mojavelinux:
The jaro_winkler dependency is a misery. We could move rubocop to a group in the gemfile that can be excluded. It only needs to run on one build anyway.

.gitattributes
```diff
@@ -0,0 +1,2 @@
...
+*.rb eol=lf
```
mojavelinux: I really dislike that fact that we have to have a .gitattributes file. Is it possible to configure these globally in GitHub Action's Windows runner?

slonopotamus: I observe exactly the same test failures on my Windows machine. Git for Windows has `core.crlf=true` by default, so all text files are checked out to working copy with CRLF line endings.

Why you're against .gitattributes? It is just a tool to configure Git behavior.

mojavelinux: Because I believe it should be configured globally. If you add CRLF newlines to source code, you are not running the same source code. Therefore, I see no reason why anyone should ever have this enabled.

slonopotamus: That's what `*.rb eol=lf` does! It says "whoever checks out *.rb files, gets them with LF line endings".

mojavelinux: To me, it's a stain on the project. I think it should be documented in the README instead. If you want to develop on this project, you can't have that setting. It may be a feature of git, but it should not be used IMO.

slonopotamus: As I already said, it is current [default system-wide setting in Git for Windows](https://github.com/git-for-windows/build-extra/blob/master/git-extra/git-extra.install#L9). And there are cases where this is a desired behavior.

I don't understand why you want people to perform manual actions to do what can be achieved with a two-line config file.

slonopotamus:
Current state: everything passes except 5 tests on JRuby+Windows: https://github.com/slonopotamus/asciidoctor-pdf/runs/409376769?check_suite_focus=true#step:5:13

I think I'll split this PR since test fixes are not GH-Actions specific and can also be reproduced outside of it.

.gitattributes
mojavelinux: Because first of all, it's not complete. If we were going to do it, we need to address every file the repository. And frankly, that's just a pain. I'll keep saying it. There's no reason to do endline conversion in a development environment. We don't reencode from UTF-8 to other encodings automatically. So why would we touch endlines. That is not a problem the project should have to deal with. That is the responsibility of the developer.

slonopotamus: I split eol issue into #1523.

slonopotamus:
This PR depends on #1523.

slonopotamus:
Fresh run: only [asciidoctor-pdf redirection should be able to write output to file via stdout](https://github.com/slonopotamus/asciidoctor-pdf/runs/409452158?check_suite_focus=true#step:5:9) fails.

slonopotamus:
Latest run: [fully passes](https://github.com/slonopotamus/asciidoctor-pdf/runs/409525684?check_suite_focus=true#step:5:8)

mojavelinux:
I think all the prerequisites are now covered.

For now, I'd like to only use GitHub Actions for what Travis doesn't already cover (basically to replace AppVeyor). So that would be Windows only.

Another possibility is to run 6 combinations.

* JRuby on Windows
* JRuby on Linux
* 2.4 on Windows
* 2.4 on Linux
* 2.7 on Windows
* 2.7 on Linux

That way we are touching all the boundaries. The more fine-grained testing of Ruby versions happens in Travis.

slonopotamus:
Matrix can be adjusted as you want it to be. Changed to Win-only.

slonopotamus:
No, wait. GH-Actions runs 922 tests, while AppVeyor runs 934. Some are disabled on Ruby-2.7?

mojavelinux:
You need to enable some environment variables:

https://github.com/asciidoctor/asciidoctor-pdf/blob/master/.travis.yml#L11-L12

Note that rghost won't likely work on Windows.

slonopotamus:
> Note that rghost won't likely work on Windows.

You can install Ghostscript, but rghost needs some setup in order to find it. Possibly, by adding a env var that would set `RGhost::Config::GS[:path]`.
```
      # Ghostscript will be available at C:\Program Files\gs\gs9.50\bin\gswin64.exe
      - name: Set up Ghostscript
        uses: crazy-max/ghaction-chocolatey@v1.0.2
        with:
          args: install --no-progress ghostscript
```

mojavelinux:
I'm ready to move forward with this. I'll probably continue to tweak it. Thanks for getting it started!

slonopotamus:
Nooooooo! :D You also hit asciidoctor/asciidoctor-epub3#253 I didn't notice that last run that triggered after I enabled pygments.rb didn't actually finish but is [still running](https://github.com/slonopotamus/asciidoctor-pdf/actions/runs/32075064).

slonopotamus:
Created #1525 to disable pygments.rb on JRuby+Windows.

mojavelinux:
No problem. I'll just fix in master.

Issue #1500, opened 2020-01-21
Travis CI cache is reuploaded for every build


slonopotamus:
For example: https://travis-ci.org/asciidoctor/asciidoctor-pdf/jobs/637866445#L655

```
store build cache
$ bundle clean
changes detected (content changed, file is created, or file is deleted):
/home/travis/build/asciidoctor/asciidoctor-pdf/.bundle/gems/ruby/2.5.0/bundler/gems/prawn-table-515f2db29486/.git/index
/home/travis/build/asciidoctor/asciidoctor-pdf/.bundle/gems/ruby/2.5.0/bundler/gems/prawn-table-515f2db29486/.git/logs/HEAD
changes detected, packing new archive
uploading master/cache--linux-trusty-e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855--rvm-2.5.7--gemfile-Gemfile.tgz
cache uploaded
```

I believe this happens for every build.

mojavelinux:
Yes, my understand is that's just how it works. Do you have a proposal to modify the behavior?

slonopotamus:
This costs 25-30s per each job on Travis currently. And reupload happens because of that prawn-table file. If that single file didn't change, cache would not be reuploaded, because your dependencies didn't change since previous run.

slonopotamus:
Can you download prawn-table through more conventional means, like rubygems?

slonopotamus:
Here's an example (from a different project) of build that does not reupload anything: https://travis-ci.org/slonopotamus/kindlegen/jobs/639359718#L228

mojavelinux:
> Can you download prawn-table through more conventional means, like rubygems?

mojavelinux:
We might be able to remove the .git directory inside of bundler's cache before the build is done so that it doesn't think it has changed.

Honestly, though, I really don't care that much that it adds half a minute since the JRuby build takes so much time anyway. That's really the bottleneck.

mojavelinux:
I may need to make follow-up commits. I'll post once it's all working.

slonopotamus:
Yep, you need to build at least twice on the same branch to see that it is really fixed.

mojavelinux:
```
nothing changed
```

I think we got it.

PR #1501, opened 2020-01-21
resolves #1503 improve Travis CI build times

### LINKED ISSUE #1503: Make CI builds on Travis run faster
0. Each of JRuby jobs takes ~12min.
1. `bundle exec rake lint` runs in every job (including JRuby). On JRuby, it takes 2min 50s. Suggested solution: only run lint on one platform.
2. `JRUBY_OPTS='--dev'` makes tests run *slower*. With it: [5min 50s](https://travis-ci.org/asciidoctor/asciidoctor-pdf/jobs/640321762#L680). Without it: [4min 50s](https://travis-ci.org/asciidoctor/asciidoctor-pdf/jobs/640327161#L685). Suggested solution: remove `JRUBY_OPTS`.
3. It is a common situation when all other jobs have finished, but JRuby is still running, leading to increased wallclock time of a build. This happens because JRuby jobs are near the end of job + the fact that number of jobs is greater than number of runners. Suggested solution: move JRuby jobs to the beginning of job list.

slonopotamus:
Note: I am unable to test that release deployment still works.

Wallclock time (with 3 parallel runners) before: ~22min
Wallclock time (with 3 parallel runners) after: ~17min

Jobs are intentionally ordered in such a way that JRuby is in the beginning. It prevents situations when all other jobs have finished, but we're still waiting for JRuby.

Some further improvements could possibly be achieved by using more modern JVM (11 instead of 8) for JRuby.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,9 +17,7 @@ rvm:
 matrix:
   include:
   - rvm: jruby-9.2.9.0
-    env: JRUBY_OPTS='--dev'
   - rvm: jruby-9.1.17.0
-    env: JRUBY_OPTS='--dev'
   - rvm: *oldest_ruby
     env: ASCIIDOCTOR_VERSION=1.5.3
   - rvm: *oldest_ruby
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,33 +8,36 @@ git:
   # use depth 5 to leave enough room for concurrent builds
   depth: 5
 language: ruby
-rvm:
-- &release_ruby 2.7.0
-- 2.6.5
-- 2.5.7
-- 2.4.9
-- &oldest_ruby 2.3.8
-matrix:
+
+jobs:
   include:
   - rvm: jruby-9.2.9.0
   - rvm: jruby-9.1.17.0
+  - rvm: &release_ruby 2.7.0
+  - rvm: 2.6.5
+  - rvm: 2.5.7
+  - rvm: 2.4.9
+  - rvm: &oldest_ruby 2.3.8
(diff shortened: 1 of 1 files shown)


mojavelinux:
Neat idea to reorder the Rubies.

Why are you defining the script for each entry in the matrix?

slonopotamus:
> Neat idea to reorder the Rubies.

If whole build is aborted when any job fails, you may also want to put Lint as number 3, because it has chances to abort _fast_ and release workers.

> Why are you defining the script for each entry in the matrix?

Absolutely no idea, I was possibly thinking about something else. _Now_ I think that it doesn't make sense and script shoould just be defined at root level.

slonopotamus:
> Why are you defining the script for each entry in the matrix?

Updated PR.

mojavelinux:
Can you remove the reordering of top-level elements so it's easier to see what is being changed?

slonopotamus:
Okay. I'll also drop a fix for #1500 because #1502 is much nicer than what I am doing here.

slonopotamus:
> Can you remove the reordering of top-level elements so it's easier to see what is being changed?

Done

mojavelinux:
It seems like the focus of this issue is switching to the jobs infrastructure to give us control over which scripts are run per matrix entry. I'm fine with that change, but we need an issue to track this work. Can you file one?

slonopotamus:
See #1503.

mojavelinux:
This looks good, thanks! I'll likely follow up with a few tweaks of my own now that you got me started.

PR #1504, opened 2020-01-22
resolves #1500 disable Travis CI cache

### LINKED ISSUE #1500: Travis CI cache is reuploaded for every build

slonopotamus:
Because it doesn't make builds faster.

----

Do not trust what this PR message says, it is just a theory. Check build timings.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,6 @@ addons:
     packages:
     - ghostscript
     - poppler-utils
-cache:
-  bundler: true
-  directories:
-  - $HOME/.rvm
 git:
   # use depth 5 to leave enough room for concurrent builds
   depth: 5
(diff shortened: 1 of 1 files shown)


slonopotamus:
Reference timings (before #1500 was fixed): https://github.com/asciidoctor/asciidoctor-pdf/issues/1503#issuecomment-577074780

slonopotamus:
Deciding how useful Travis cache is for asciidoctor-pdf is left as an exercise to the reader.

mojavelinux:
If we don't see gains, we shouldn't use it. I actually prefer not to have a cache because then the run is cleaner. It was my understanding it would be more efficient, but that turns out not to be the case. So I say we just drop it.

PR #1511, opened 2020-01-23
messing around with travis jobs

mojavelinux:


.travis.yml
```diff
@@ -5,38 +5,36 @@ addons:
     - ghostscript
     - poppler-utils
 git:
-  # use depth 5 to leave enough room for concurrent builds
-  depth: 5
+  depth: 5 # use depth 5 to leave enough room for concurrent builds
```
slonopotamus: Does this setting actually do anything useful? I've never set this in other Travis scripts.

mojavelinux: Yes, it tells get to only fetch index packs for files present in recent commits. The larger the repo, and the more binaries it has, the bigger the difference. You can try it with `git clone --depth 5 <url>`. (2.5s vs 9s for asciidoctor-pdf)

mojavelinux:
Done messing around.

PR #1512, opened 2020-01-23
upgrade CI dist to bionic

mojavelinux:


slonopotamus:
> use 2.6.5 as the default Ruby since it's pre-installed

[2.7 also installs in no time](https://travis-ci.org/asciidoctor/asciidoctor-pdf/jobs/640846074#L280), possibly [docs](https://docs.travis-ci.com/user/reference/bionic/#ruby-support) are outdated a bit.

mojavelinux:
Yeah!

mojavelinux:
However, it seems like JRuby doesn't like bionic.

mojavelinux:
I decided to take your advice and drop the Lint stage. I was able to name the first job in the Test stage, which is good enough for me.

mojavelinux:
We have to run JRuby tests on trusty still.

mojavelinux:
Lookin' good at 12min.

Issue #1518, opened 2020-01-24
pdf-page-size attribute not working


janicemhdc:
I placed the following at the top of my main adoc file:

`:pdf-page-size: 11in x 14in`

which, according to the PDF Theming Guide, should work as a document attribute:

```
pdf-page-size | Named size \| Measurement[width, height] | :pdf-page-size: 6in x 9in
```

(see https://github.com/asciidoctor/asciidoctor-pdf/blob/master/docs/theming-guide.adoc#theme-related-document-attributes)

But it has no effect at all on my PDF output. It's still generating a standard 8.5x11 page

rillbert:
Could it be that you need to have the measures in the same format as the pdf-page-margin attribute, that is:

`:pdf-page-size: [11in, 14in]`

and that the example in the guide has a typo?

/Anders

janicemhdc:
Yes, that's it! Thanks! I'll submit a doc PR with the update

mojavelinux:
The original example is valid syntax. There's even a test to verify it. https://github.com/asciidoctor/asciidoctor-pdf/blob/master/spec/page_spec.rb#L45-L53 Are you using the 1.5.0.rc.2 release?

janicemhdc:
asciidoctor-pdf --version
Asciidoctor PDF 1.5.0.beta.8 using Asciidoctor 2.0.10 [https://asciidoctor.org]
Runtime Environment (ruby 2.6.3p62 (2019-04-16 revision 67580) [x86_64-darwin17])

janicemhdc:
Assuming the test is run in isolation, I respectfully submit that it's not a valid test given that it's setting the size to the default used if no size is specified.

mojavelinux:
Please use 1.5.0.rc.2. It could be that this was fixed in one of the recent releases.

mojavelinux:
> it may not be a valid test given that it's setting the size to the default used if no size is specified.

The default if not size is specified is A4. The test is changing it to Letter. So the test is valid.

mojavelinux:
I checked and this works even on beta.8. So there must be something incorrect / incompatible about your document. Could you share a complete sample that demonstrates the issue?

mojavelinux:
We need to figure out what the scenario is so that we can add a better test, if necessary ;)

janicemhdc:
[...]

PR #1519, opened 2020-01-24
Fix bad example per issue #1518

janicemhdc:
The pdf-page-size example in the current docs is incorrect. This update fixes it.


mojavelinux:
Although the proposed syntax is correct, either syntax is valid.

mojavelinux:
We've worked out that both are valid. However, to be consistent with the other example in the table, I've decide to move forward with this change.

PR #1523, opened 2020-01-26
Disable Git EOL autoconversion on Windows

slonopotamus:
By default ([1]), Git for Windows performs automatic LF->CRLF conversion for text files
when checks them out into working copy.

While this behavior can be desired sometimes, it leads to issues in asciidoctor-pdf:

1. Some PDF files get corrupt, for example examples/basic-example.pdf
2. JRuby doesn't properly parse inline string blocks in Ruby files when they have CRLF line endings

Given that user cannot easily disable CRLF conversion specifically for asciidoctor-pdf repo
when they do `git clone`, this commit completely disables CRLF conversion via .gitattributes.

So, files in working copy will be byte-to-byte identical no matter what platform or environment is used
to checkout asciidoctor-pdf repository.

[1]: https://github.com/git-for-windows/build-extra/blob/master/git-extra/git-extra.install#L9

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,7 @@
+/*file eol=lf
+/bin/* eol=lf
+*.adoc eol=lf
+*.pdf -text
+*.rb eol=lf
+*.svg eol=lf
+*.yml eol=lf


slonopotamus:
Git EOL handling docs: https://git-scm.com/docs/gitattributes#_code_text_code

slonopotamus:
I personally would instead just do

```
*.rb eol=lf
*.pdf -text
```

Because automatic CRLF feature has at least one good effect: it prevents CRLF from entering repository by converting them to LF on commit. With auto-CRLF disabled, Windows users might start sending you PRs where files (including Ruby files) will contain CRLF line endings, simply because they created them in an editor that writes CRLF (Windows default line endings).

mojavelinux:
Thank you for this proposal. While I have voiced my disdain for the need for this file, I'm also tired of wasting time on it. So I'm going to go ahead an accept it so we can move on to more important matters.

Before I do so, I'd like to switch to a more complete version of your original suggestion:

```
/*file eol=lf
/bin/* eol=lf
*.adoc eol=lf
*.pdf -text
*.rb eol=lf
*.svg eol=lf
*.yml eol=lf
```

Can you update the PR and also verify that this hits all the right files? If you do so, I'll merge it.

slonopotamus:
You forgot about `*.pdf -text`

mojavelinux:
;)

mojavelinux:
Thanks for updating!

PR #1525, opened 2020-01-28
resolves #1520 disable pygments.rb on JRuby on Windows because it hangs

### LINKED ISSUE #1520: Add CI testing for JRuby on Windows
Subj. This is a very fragile environment with its own class of bugs that deserves dedicated testing.

slonopotamus:
See https://github.com/asciidoctor/asciidoctor-epub3/issues/253 for more details.

### DIFF
--- a/lib/asciidoctor/pdf/converter.rb
+++ b/lib/asciidoctor/pdf/converter.rb
@@ -1621,7 +1621,8 @@ module Asciidoctor
             end
           when 'pygments'
             unless defined? ::Pygments::Ext::BlockStyles
-              highlighter = nil if (Helpers.require_library PygmentsRequirePath, 'pygments.rb', :warn).nil?
+              # pygments.rb hangs on JRuby for Windows, see https://github.com/asciidoctor/asciidoctor-epub3/issues/253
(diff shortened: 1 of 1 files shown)


mojavelinux:
I not sure we need this logic in Asciidoctor PDF because Asciidoctor PDF does not load Pygments automatically like Asciidoctor EPUB3. The user just shouldn't enable pygments when running on that platform. Though I can see the point of "protect the user". Even then, it's a bit more complicated because it's core that loads Pygments since Asciidoctor 2.

slonopotamus:
If you don't autoload, then no need to protect. Or at least this kind of protection (or, even explicit disabling of pygments.rb on problematic platform) should be in Asciidoctor core instead. But you do autoload currently.

mojavelinux:
No, we only autoload if syntax-highlighter is set to pygments. And by default, it is not set.

mojavelinux:
I don't want to block usage because when (if) pygments is fixed, then we have a gem out there that is blocking it from being used. So it should only be disable in the test suite. (Asciidoctor EPUB3 is different since it automatically activates pygments by default).

slonopotamus:
Makes sense.

PR #1531, opened 2020-01-30
Windows Actions CI - use $ErrorActionPreference for bundler install step

MSP-Greg:
At present, all Actions Windows Rubies have Bundler installed.  If an update is needed, doing so with `gem update --system` is probably a better idea.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,16 +12,16 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
       - name: Install Ruby
-        uses: eregon/use-ruby-action@v1.10.0
+        uses: eregon/use-ruby-action@v1
         with:
           ruby-version: ${{ matrix.ruby }}
       - name: Install dependencies
         run: |
-          gem install -N bundler
-          # NOTE using bundle config causes bundle install to crash; see https://github.com/eregon/use-ruby-action/issues/13
-          #bundle config --local path .bundle/gems
-          #bundle config --local without docs lint coverage
-          #bundle --jobs 3 --retry 3
-          bundle --jobs 3 --retry 3 --path .bundle/gems --without docs lint coverage
+          # Fix for PowerShell issue, non default value (stop) used by Actions
+          # see https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_preference_variables?view=powershell-7#erroractionpreference
+          $ErrorActionPreference = 'Continue'
+          bundle config --local path .bundle/gems
+          bundle config --local without docs lint coverage
+          bundle --jobs 3 --retry 3
       - name: Test
         run: bundle exec ruby -w $(bundle exec ruby -e 'print File.join Gem.bindir, %q(rake)') spec


.github/workflows/ci.yml
```diff
@@ -12,16 +12,15 @@ jobs:
...
-          gem install -N bundler
-          # NOTE using bundle config causes bundle install to crash; see https://github.com/eregon/use-ruby-action/issues/13
-          #bundle config --local path .bundle/gems
-          #bundle config --local without docs lint coverage
-          #bundle --jobs 3 --retry 3
-          bundle --jobs 3 --retry 3 --path .bundle/gems --without docs lint coverage
+          bundle config --local path .bundle/gems
+          bundle config --local without docs lint coverage
+          # warning output can confuse Powershell in CI, (multi-line scripts & exit codes)
+          $env:RUBYOPT = '-W0'
```
slonopotamus: 1. What if bundler call is split into a separate step? Does it also fail to print warnings?
2. You lost `--without docs lint coverage`

slonopotamus: @mojavelinux and what's the point storing gems in `.bundle/gems`? Ain't everything supposed to work regardles of gems location? Whole CI VM will be dropped anyway after build is finished.

MSP-Greg: @slonopotamus 

> What if bundler call is split into a separate step? Does it also fail to print warnings

It's not about it failing to print warnings.  It's about warnings causing the step to fail when they shouldn't.

There's been discussion in RubyGems about this, in that warnings are printed when a user has no control over the gem in question.  The idea was to change the warnings to only show when the gem is built, not when it is installed.

slonopotamus:
Oh, you mean it exits with 1 on warnings? That's weird.

MSP-Greg:
@slonopotamus

Re gem location (`.bundle/gems`), that can be cached in Actions so they aren't downloaded all the time.  But, I think the caches empty after seven days, so if there aren't a lot of commits/PR's...

I didn't set up caching.  Using `.bundle/gems` can be helpful for people working locally.

MSP-Greg:
@slonopotamus 

> Oh, you mean it exits with 1 on warnings? That's weird.

Yes, see some of the previous actions jobs.

Quite the PITA.  I've had the issue before (and also on Appveyor).  As I mentioned in https://github.com/eregon/use-ruby-action/issues/13, I do not have the issue running the commands locally.  Someday I have to track the issue down re PowerShell configuration...

slonopotamus:
I would avoid caching unless there's a strong need to improve build times, because it can hide bugs when gems are available because they were cached and not because there's a dependency on them.

mojavelinux:
> all Actions Windows Rubies have Bundler installed.

Oh, I didn't realize that. Then no need for this step. I was just following the README https://github.com/eregon/use-ruby-action#caching-bundle-install (which I now see says that Bundler is already there).

.github/workflows/ci.yml
mojavelinux: The use of path is no about caching. It's about testing the installation of the CLI correctly and in isolation. Without specifying a path, the CLI location test has to be partially faked.

mojavelinux:
> Using .bundle/gems can be helpful for people working locally.

It also forces Bundler to install the CLI for the current gem, which then allows me to test the CLI properly.

mojavelinux:
But why do we get warnings when `bundle config` is used and not when the CLI flags like `--path` are used? I don't get it.

.github/workflows/ci.yml
mojavelinux: Sweet trick!

mojavelinux: Except don't we need to unset it afterwards so it doesn't remain set while running tests?

MSP-Greg: > Except don't we need to unset it afterwards so it doesn't remain set while running tests?

No.  Big difference between Azure Pipelines / GitHub Actions and Travis / AppVeyor.  In Actions, changes to the environment do not affect subsequent steps, unless one uses something like:
```
echo '::set-env name=JOBS::'-j$((1 + $(nproc --all)))
```

Sometimes it's helpful, often not...

mojavelinux: Thanks for the info!

mojavelinux:
So my question still stands, what is the warning? And how is this different from the warning when `--path` is used?

```
[DEPRECATED] The `--path` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set path '.bundle/gems'`, and stop using this flag
```

slonopotamus:
> But why do we get warnings when bundle config is used and not when the CLI flags like --path are used? I don't get it.

Could you once again give a link to workflow run with warnings? AFAIR, there was something about prawn-tables, but not about config options you use.

mojavelinux:
> there was something about prawn-tables

That was just the last message logged. It has nothing to do with this situation.

> But why do we get warnings when bundle config is used and not when the CLI flags like --path are used?

I've already linked to the logs. But here's an example:

https://github.com/asciidoctor/asciidoctor-pdf/runs/416845301

(view the raw log)

slonopotamus:
@mojavelinux have you splitting that in separate steps so it becomes clear which exact comand returns 1?

mojavelinux:
Here's what I'm saying. If I run this it works fine:

```
bundle --jobs 3 --retry 3 --path .bundle/gems --without docs lint coverage
```

If I run this, it fails:

```
bundle config --local path .bundle/gems
bundle config --local without docs lint coverage
bundle --jobs 3 --retry 3
```

Why does it die in one case and not in the other. It's exactly the same set of instructions. The only difference is that there are 3 commands instead of 1.

Btw, the same thing happens if I write directly to .bundle/config instead of using `bundle config`.

mojavelinux:
The command that fails is `bundle --retry 3 --jobs 3`. If we split in separate steps, maybe that also solves it. I guess we could try.

MSP-Greg:
Have at it.  As I've mentioned, I could not repo this locally.  See
https://github.com/eregon/use-ruby-action/issues/13#issuecomment-580485479

mojavelinux:
I've been able to confirm that using separate steps also solves the problem. I'm going to go with that approach instead. Thanks for helping us troubleshoot this, @MSP-Greg! Much appreciated!

MSP-Greg:
Give me a second, running CI now...

mojavelinux:
Sure thing, take your time.

MSP-Greg:
I changed it to use $ErrorActionPreference.  I'll write some code to locally see if I can determine what's going on...

mojavelinux:
But you didn't remove `$env:RUBYOPT = '-W0'`, which we already determined fixes the issue. Did you mean to leave that in?

MSP-Greg:
> Did you mean to leave that in?

I knew that.  No, I did not.   Sorry, fixed.

mojavelinux:
Yeah, that worked. Wow, now we have like 3 different ways of solving this issue.

I'll quickly summarize:

1. Run `$env:RUBYOPT = '-W0'` before calling `bundle install` if `bundle config` is used in same step
2. Run `$ErrorActionPreference = 'Continue'` before calling `bundle install` if `bundle config` is used in same step
3. Move `bundle config` to separate step

Does that look right?

MSP-Greg:
Yes.  I still can't repo this locally.  Every command seems to exiting with zero.  I hate this.

I think $ErrorActionPreference = 'Continue' is probably the best, as it shows what seems to be  the problem...

mojavelinux:
Thanks again! I'll stick with the final result then. Cheers!

slonopotamus:
> I'll stick with the final result then.

Splitting steps would keep build script cross-platform. You don't need that right now, though.

mojavelinux:
I went ahead and did both. That way we can monitor output in the interim, but disable the setting when it comes time to go cross platform.

PR #1558, opened 2020-02-15
resolves #1557 fix spacing after first line of indented paragraph

### LINKED ISSUE #1557: Prose text indent reduces spacing below first paragraph line
When the `prose-text-indent` property is set to greater than zero, the space between the first and second lines of the paragraph is reduced. The space is reduced by the same amount regardless of the indent amount. This doesn't seem intentional as nothing is mentioned in the documentation about this side-effect, and I couldn't find any properties related to setting the space either.

Source (bug.asc):

```adoc
= asciidoctor-pdf bug
:doctype: book

XXXX XXXX XXXX XXXX XXXX XXXX XXXX XXXX XXXX XXXX XXXX XXXX XXXX XXXX XXXX
XXXX XXXX XXXX XXXX XXXX XXXX XXXX XXXX XXXX XXXX XXXX XXXX XXXX XXXX XXXX
XXXX XXXX XXXX XXXX XXXX XXXX XXXX XXXX XXXX XXXX XXXX XXXX XXXX XXXX XXXX
XXXX XXXX XXXX XXXX XXXX XXXX XXXX XXXX XXXX XXXX XXXX XXXX XXXX XXXX XXXX
```

Style (bug.yml):

```yaml
page-size: Letter
prose-text-indent: 0.1in
```

Here's what it looks like rendered as a PDF:

bug-indent

Here's what it looks like rendered as a PDF after setting `prose-text-indent` to 0:

bug-no-indent

```
$ asciidoctor-pdf --version
Asciidoctor PDF 1.5.0 using Asciidoctor 2.0.10 [https://asciidoctor.org]
Runtime Environment (ruby 2.3.3p222 (2016-11-21 revision 56859) [x86_64-cygwin]) (lc:UTF-8 fs:Windows-1252 in:UTF-8 ex:UTF-8)
```

mojavelinux:


mojavelinux:
I figured it out. The value for the spacing after the first line was not being computed correctly. It affected some fonts, but not others. The fix will ensure the value is computed correctly.

PR #1565, opened 2020-02-20
fix CI failure and warnings running on JRuby 9.2

mojavelinux:


mojavelinux:
Travis was configured to use the --client flag, which is deprecated.

For some reason, the script to resolve the rake path was not working, but found a simpler way.

PR #1601, opened 2020-03-11
relax dependency on concurrent-ruby

pravi:
concurrent-ruby is compliant with semantic versioning https://github.com/ruby-concurrency/concurrent-ruby#versioning so ~> 1.1 is enough


mojavelinux:
Seems reasonable to me. Thanks!

PR #1652, opened 2020-04-18
Handle rename of eregon/use-ruby-action to ruby/setup-ruby

slonopotamus:
While action is available via both names, former one is deprecated.


mojavelinux:
Thanks for keeping this updated @slonopotamus!

PR #1686, opened 2020-05-08
Include reference to KDP and their need for flattened, i.e. optimized files.

### LINKED ISSUE #1684: Add "flattening" step to PDF production
I don't really know anything about PDF generating engines, but is it possible to ask the engine to "flatten" the final PDF? Or at least make that a configurable option?

I found a thread that sounded like a similar issue, and added my own bit to it for community contribution => https://discuss.asciidoctor.org/asciidoctor-pdf-style-for-KDP-print-book-td7789.html#a7978

Having the tools offer this would be nice.

gregturn:
Resolves #1684.


mojavelinux:
Thanks @gregturn!

PR #1718, opened 2020-06-23
resolves #1715 use more specific regexp for hyphenation

### LINKED ISSUE #1715: Hyphenation exceptions not always respected
Some language files of `Text::Hyphen` contain exceptions for some words to explicitly specify their hyphenation (e.g. the [Dutch one](https://github.com/halostatue/text-hyphen/blob/master/lib/text/hyphen/language/1.9/nl.rb#L1174)). However, these are not respected if that word is right next to another character (e.g. punctuation).

Example:
Comparison of hyphenation without and with adjacent character
The allowed hyphenation according to the exception is "sou-ve-niertjes".

This is because `hyphenate_words` matches on any non-whitespace character and passes the resulting tokens to the hyphenator, which then uses it for a lookup in the exceptions hash and that fails if the token includes an adjacent character.

This could be fixed by using a narrower regexp for the tokenization, e.g. `[[:alpha:]]` or `[[:word:]]`. Especially the latter seems to be exactly intended for such purposes and worked well in my tests. Do you see any problem with that? Maybe breaking the hyphenation in some special cases? Additionally, this would avoid the special handling of hyphenated hyphens, since the two words will be hyphenated separately.

klonfish:
This resolves #1715 by using the `[[:word:]]` character class instead of all non-whitespace characters for tokenization.

I decided to use unit tests for this instead of integration tests, since that gave me more control over the exceptions (such that it is independent of possible changes in the language files) and made it easier to check the output. However, as I'm completely new to Ruby, the implementation may not conform to some best practices. So please let me know if there is something that can be improved.

### DIFF
--- a/lib/asciidoctor/pdf/text_transformer.rb
+++ b/lib/asciidoctor/pdf/text_transformer.rb
@@ -6,10 +6,10 @@ module Asciidoctor
       XMLMarkupRx = /&#?[a-z\d]+;|</
       PCDATAFilterRx = /(&#?[a-z\d]+;|<[^>]+>)|([^&<]+)/
       TagFilterRx = /(<[^>]+>)|([^<]+)/
-      WordRx = /\S+/
+      ContiguousCharsRx = /\p{Graph}+/
+      WordRx = /\p{Word}+/
       Hyphen = '-'
       SoftHyphen = ?\u00ad
-      HyphenatedHyphen = '-' + SoftHyphen
 
       def capitalize_words_pcdata string
         if XMLMarkupRx.match? string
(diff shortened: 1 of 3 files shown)


lib/asciidoctor/pdf/text_transformer.rb
```diff
@@ -6,7 +6,7 @@ module TextTransformer
       XMLMarkupRx = /&#?[a-z\d]+;|</
       PCDATAFilterRx = /(&#?[a-z\d]+;|<[^>]+>)|([^&<]+)/
       TagFilterRx = /(<[^>]+>)|([^<]+)/
-      WordRx = /\S+/
+      WordRx = /[[:word:]]+/
```
mojavelinux, line 9: In Asciidoctor, we use the character properties instead of the POSIX character classes since the former are Unicode-friendly. Therefore, I will change this to `\p{Word}`.

mojavelinux:
This is a solid change. I'm going to take a slightly different approach to the tests because we want to preserve the assertion that the text-hyphen gem is autoloaded.

PR #1721, opened 2020-07-03
Update README.adoc

djencks:
keep together is not documented anywhere, but it's easy to use once you discover how.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -84,7 +84,9 @@ But don't miss the <<Highlights>> to get a preview of what's possible.
 * Page numbering
 * Double-sided (aka prepress) printing mode (i.e., margins alternate on recto and verso pages)
 * Customizable running content (header and footer)
-* “Keep together” blocks such as open blocks specified with the "unbreakable" option `[%unbreakable]` (i.e., page breaks avoided in certain block content)
+* “Keep together” blocks (i.e., page breaks avoided in certain block content):
+** Explicitly delimited blocks other than open blocks, including tables and lists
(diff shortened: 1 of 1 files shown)


mojavelinux:
I agree we need to draw attention to this feature, but this change removes mention of the default keep together behavior for all other blocks. In other words, open blocks are not keep together by default, requiring the unbreakable option to be set. All other delimited blocks keep together by default.

mojavelinux:
This still isn't correct. Only open blocks support the unbreakable option. All other delimited blocks are implicitly unbreakable. I think these should be two separate points.

djencks:
My impression is that paragraph blocks and section blocks are not unbreakable, so your claim about "the default keep together behavior for all other blocks" isn't accurate either.  Is it explicitly delimited blocks + open blocks with [%unbreakable] that are unbreakable?

mojavelinux:
I said all other **delimited** blocks. That excludes sections and paragraphs.

mojavelinux:
>  Is it explicitly delimited blocks + open blocks with [%unbreakable] that are unbreakable?

Correct.

djencks:
I quoted your first comment.

djencks:
Are tables and lists unbreakable?

mojavelinux:
Lists no.

Tables are a bit of a hybrid, and all behavior is controlled by prawn-table. I think cells try to stay together, but there are numerous edge cases I haven't explored.

djencks:
based on how paged.js does, paginating tables properly is very difficult!

mojavelinux:
This is great. Thanks!

PR #1766, opened 2020-09-05
resolves #1763 add support for after-toc keyword to start-at keys for page numbering and running content

### LINKED ISSUE #1763: Add support for the after-toc keyword for start-at keys in theme
If you want to start the page numbering or running content after the toc in a book (doctype=book), and the toc is in the default location, you can use the keyword "body". However, if the toc is placed after the preamble, or in a custom location using the toc macro, you after to specify a body offset (integer value). However, you don't know this offset until the document is converted.

To simplify this configuration, it should be possible to specify the keyword value "after-toc". When this value is specified, the converter will compute the body offset and start the page numbering and running content after it.

The advantage of this keyword is that this will work regardless of the content in the document.

mojavelinux:


Leon0402:
Hi, 

I've tested it with a preamble and a toc over two sites and in works well, thank you!

There are a few more optional configurable things I could think of, which would be nice to have: 

- Also don't count these pages (e.g. the page after the toc is page 1 and not whatever real page it is)
- Count the preamble + toc as well, but separately with roman numbers (this makes only sense in conjunction with the first point) -> This is what I have seen most so far in documents

mojavelinux:
Thanks for testing!

> Also don't count these pages (e.g. the page after the toc is page 1 and not whatever real page it is)

That's exactly what this PR does. You'd use the following theme configuration:

```
extends: default
page-numbering-start-at: after-toc
running-content-start-at: after-toc
```

Or, to hide the running content just for the toc pages, you can use:

```
extends: default
page-numbering-start-at: after-toc
```

and then in your document:

```
[%noheader%nofooter]
toc::[]
```

> Count the preamble + toc as well, but separately with roman numbers (this makes only sense in conjunction with the first point) 

Again, that's what this pull request does.

Every page has to have a number. The question is whether it's a roman numeral or an integer. And wherever the integers start, they increment sequentially for the remainder of the document.

Leon0402:
Hi, 

thanks for the instructions! I actually did it wrong the first time and used just "running-content-start-at: after-toc", my bad. 
You are right, the PR is covering everything needed, thanks a lot!

mojavelinux:
Fantastic! Thanks for testing!

PR #1797, opened 2020-10-19
resolves #1699 add default themes optimized for print

### LINKED ISSUE #1699: For media=prepress switch base font color to black
Asciidoctor PDF uses Gray 20 (#333333), a slightly lighter-than-black as it’s base-font-color due to my references when reading a PDF on a device. 

When you’re actually going to press, you need “real” black (#000000) to avoid a faded look. 

Adjusting media=prepress to that would reduce the odds of someone sending a prepress rendered PDF KDP Print (or any other POD service) and be surprised/disgusted with the results. 

Additionally this should be mentioned in the docs.

mojavelinux:


mojavelinux:
@gregturn Let me know what you think about this idea.

gregturn:
Looks like a great approach. 

I’ll definitely give a test run when it’s time to turn the crank on my next book

mojavelinux:
Excellent!

This nice part about this approach is that we can really fine-tune the theme so that we're producing the best PDF we can for print.

thucke:
+1

PR #1809, opened 2020-10-31
Add Prawn 2.3 to the build matrix

pevik:
Suggested-by: Dan Allen <dan.j.allen@gmail.com>
Signed-off-by: Petr Vorel <petr.vorel@gmail.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,6 +49,8 @@ jobs:
     - bundle exec ruby scripts/switch-to-prawn-master.rb
     - rm -f Gemfile.lock
     - bundle --jobs 3 --retry 3
+  - rvm: *default_ruby
+    env: PRAWN_VERSION='~> 2.3'
   - stage: release gem
     rvm: *default_ruby
     script: skip


mojavelinux:
The build failure is not related to this change. However, it made me realize that we're already testing against the master branch of Prawn (which is 2.3.0 with some additional changes). I had fixed these changes on master, but now I see those fixes need to go into the v1.5.x branch too.

pevik:
Yes, I wondered, if you really wanted to test both prawn master and 2.3 (yes, the difference on both are currently small). Feel free to close this.
I' more concerned about adding 2.3 support in `asciidoctor-pdf.gemspec` (and releasing this change).
And thanks for backporting prawn master fixes to v1.5.x.

mojavelinux:
I don't really have anything against testing 2.3 exactly. I just wanted you to know that technically we have already verified it (as it has been well tested throughout the Prawn 2.3 release cycle).

I noticed while applying the patch that the version of pdf-core we've been testing is not correct, so I'm working on fixing that now.

pevik:
Right, I rebased the branch then.

mojavelinux:
Thanks!

PR #1819, opened 2020-11-12
speedup JRuby CI

slonopotamus:
Before: [9m2s](https://github.com/asciidoctor/asciidoctor-pdf/runs/1389540113?check_suite_focus=true#step:11:30)
After: [7m5s](https://github.com/slonopotamus/asciidoctor-pdf/runs/1389664290?check_suite_focus=true#step:11:30)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,7 +57,6 @@ jobs:
           primary: true
     runs-on: ${{ matrix.os }}
     env:
-      JRUBY_OPTS: '-J-XX:+TieredCompilation -J-XX:TieredStopAtLevel=1 -J-Xss2m -Xcompile.invokedynamic=false'
       MENTOS_TIMEOUT: 30 # give pygments.rb calls more time to execute (in seconds)
       PYGMENTS_VERSION: '~> 1.2.0' # test runner will automatically disable Pygments tests on JRuby for Windows
       RGHOST_VERSION: '0.9.7'


mojavelinux:
Hmm, the point of the opts was to speed things up, not to slow things down. I think those were options I had been using to speed up startup time. But at the cost of longer runs like the test suite. It's probably best to give us what JRuby wants to give us by default and trust in them ;)

slonopotamus:
Hmm... Either GH-Actions runners are not all born equal or there's something severely affecting execution times. You had [6m37s](https://github.com/asciidoctor/asciidoctor-pdf/runs/1389429787?check_suite_focus=true#step:10:30) even without these changes.

mojavelinux:
I'll run it a few times and see whether we can detect an improvement. It could be that it just doesn't make much of a different either way, which would be an argument to just keep it simple and not specify any opts.

mojavelinux:
Here are the opts JRuby uses for it's own tests. https://github.com/jruby/jruby/blob/master/.travis.yml#L37 Though that doesn't necessarily mean they are doing it to make the tests run faster.

slonopotamus:
Travis is more memory-constrained than GitHub Actions, JRuby definitely targets low memory usage, not execution time:

https://github.com/jruby/jruby/commit/7a73edd2774bb6fe131f16db9e7306d15eac3189
https://github.com/jruby/jruby/commit/20f0a923872848284a831ca04d0555cd9cb5b122
https://github.com/jruby/jruby/commit/696656709875c0879b67f81305999ed9dd94c48d

And addition of `-XX:+TieredCompilation -XX:TieredStopAtLevel=1` was done back in 2014 (https://github.com/jruby/jruby/commit/dd9b0c1ee166994db412ed4769881b9def27f2cf). It could stop being effective for newer JVMs or for different load.

mojavelinux:
It looks like these settings have little to no effect anyway, so let's just keep it simple and remove them.

PR #1843, opened 2020-12-17
Drop workaround for Bundler 2.2.0 on JRuby on Windows bug

slonopotamus:
It was fixed in Bundler 2.2.1

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,11 +67,6 @@ jobs:
       uses: ruby/setup-ruby@v1
       with:
         ruby-version: ${{ matrix.ruby }}
-    - name: Reinstall Bundler
-      if: matrix.os == 'windows-latest' && matrix.ruby == 'jruby-9.2'
-      run: |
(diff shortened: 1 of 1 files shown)


slonopotamus:
Err... Wut.

slonopotamus:
I believe that build failure is not caused by my changes.

mojavelinux:
It's possible it's due to a change in Bundler. It seems to be having difficulty resolving the prawn-table gem from git. Perhaps kick the build again?

slonopotamus:
Hmm... Good point. Seems like I cannot convert PR into a WIP after I created it normally. Pretend that it is WIP starting from NOW :D

slonopotamus:
Okay, seems like Bundler 2.2.x is still a bit broken on JRuby + Windows. Need to bugreport it to Bundler.

slonopotamus:
Reported: rubygems/rubygems#4195

mojavelinux:
Thank you for your diligence in following up on this. :trophy:

PR #1850, opened 2020-12-31
add Ruby 3 to CI matrix; make it primary; patch bug in Prawn; use it for releases

mojavelinux:


mojavelinux:
what?

mojavelinux:
@slonopotamus I have absolutely no idea why this is failing on Ruby 3. I can run the tests fine on Ruby 3 on my own machine and even in a Docker container using the Ruby installed by ruby-setup.

mojavelinux:
@eregon Any idea why this might be failing on Ruby 3? Ruby 2.7 works just fine in GitHub Actions, and I can run the tests on my own machine. But it cannot seem to find core gems in GitHub Actions on Ruby 3.

eregon:
Mmh, rexml is a bundled gem according to https://www.ruby-lang.org/en/news/2020/12/25/ruby-3-0-0-released/ (it was stdlib before).
That means, if you use Bundler, you need to depend on it explicitly for Ruby 3+.

mojavelinux:
@erebor Aha! I did not make that connection. Thanks for the insight! I'll try to add it as a direct dependency and see how it goes.

mojavelinux:
It looks like that worked!

(I still don't quite understand why this is necessary, but we'll cross the bridge we have to cross if users still have issues).

PR #1852, opened 2021-01-01
upgrade to Prawn 2.4

mojavelinux:
- allows gem to be installed on Ruby 3


mojavelinux:
Until we figure out what's going on with #1850, we'll at least upgrade to prawn 2.4 with the existing set of Rubies.

PR #1874, opened 2021-02-02
resolves #1875 allow vertical lines in table grid to be styled separately

### LINKED ISSUE #1875: Allow vertical and horizontal lines of table grid to be styled independently
Currently, the vertical and horizontal lines of the table grid can only be controlled uniformly (same color, width, and style). Allow the theme to control the properties of each axis independently.

Hextremist:
Hi,

I need to style the grid inside tables differently for the horizontal and vertical lines.  This pull request introduces three new theme settings:
* table-grid-color-vertical
* table-grid-style-vertical
* table-grid-width-vertical

If any of them are used, the corresponding settings (e.g. table-grid-style) is used horizontally only.


mojavelinux:
I'm open to the idea, but I don't want to introduce dedicated keys for this. That just makes the configuration too busy. What I would rather do is honor a matrix value, where the first entry is the horizontal value and the second is the vertical value. That then matches what the property accepts anyway. And we can coerce a single value to a matrix value.

Hextremist:
OK, I think I understand.  Something like:
  
table-grid-width: [0.3,1.5]

right?

mojavelinux:
Precisely.

Hextremist:
I am mostly done, but there is a slight problem here related to the "css" parsing of the colors. The width and style works fine.

table:
  grid-color: *????*
  grid-width: [1,2]
  grid-style: [dashed,solid]
  
I have tried some different values here, output of `p table_grid_color` on the right:

[00ff00,ff0000] => ["00FF00", "00FF00"]
[#00ff00, #ff0000] => ["DDDDDD", "DDDDDD"]
["00ff00","ff0000"] => ["00FF00", "00FF00"]
["#00ff00","#ff0000"] => ["#00FF0", "#00FF0"]

None of these resulted in what I wanted.  Any ideas?

Hextremist:
By the way, can I run the lint test locally before pushing somehow?

mojavelinux:
Keys that end with -color undergo a translation from different color formats. You can find the logic here: https://github.com/asciidoctor/asciidoctor-pdf/blob/master/lib/asciidoctor/pdf/theme_loader.rb#L252. This is presenting a case that has never been considered. So we'll have to find a solution.

mojavelinux:
I think what we'll need to do is add a new case like this:

```rb
when 2
  return value.map {|it| to_color it }
```

It's a bit weird since now we are allowing an array which itself has a fixed number of colors. But I can't think of another solution at the moment.

mojavelinux:
The other solution would be to add the exception where to_color is called and detect the 2-element array there instead. That way, to_color doesn't have to know about it. And we already have other exceptions. See https://github.com/asciidoctor/asciidoctor-pdf/blob/master/lib/asciidoctor/pdf/theme_loader.rb#L164

mojavelinux:
> By the way, can I run the lint test locally before pushing somehow?

bundle exec rake lint

I just realized this isn't mentioned in the contributing document. I'll add it.

mojavelinux:
I have created issue #1875 to track this enhancement. In the future, always create an issue for a change request so that we can use the issue tracker and CHANGELOG to track it. You can then state in the issue that you indent to submit a PR. The discussion about the change should primarily be in the issue, whereas the discussion in the PR is reserved for code review. Thanks!

Hextremist:
> I have created issue #1875 to track this enhancement. In the future, always create an issue for a change request so that we can use the issue tracker and CHANGELOG to track it. You can then state in the issue that you indend to submit a PR. The discussion about the change should primarily be in the issue, whereas the discussion in the PR is reserved for code review. Thanks!

OK, sorry about that.  I have done some RTFM here, so hopefully things will be better the next time.  Since this is  pure rubycop/code comment:

```
lib/asciidoctor/pdf/converter.rb:2194:11: C: [Correctable] Style/IfUnlessModifier: Favor modifier if usage when having a single-line body. Another good alternative is the usage of control flow &&/||.
          if table_grid_width.size != 2 || table_grid_width.detect {|x| !x.is_a? Numeric }
          ^^
```

Can't say I prefer 
```
raise 'table-grid-width array must contain two numbers separated by comma, e.g [0.5,1]' if table_grid_width.size != 2 || table_grid_width.detect {|x| !x.is_a? Numeric }
```
over
```
if table_grid_width.size != 2 || table_grid_width.detect {|x| !x.is_a? Numeric }
  raise 'table-grid-width array must contain two numbers separated by comma, e.g [0.5,1]'
end
```
but the cop is the boss...

mojavelinux:
> but the cop is the boss...

yeah, I'm having difficulty tuning that cop. I think the problem is that I haven't set the max line width, so it ends up always preferring the postfix conditional. What I really want is the postfix conditional for short lines, but for long lines the block syntax. We'll get there.

lib/asciidoctor/pdf/theme_loader.rb
```diff
@@ -163,7 +163,12 @@ def process_entry key, val, data, normalize_key = false
           end
         elsif key.end_with? '_color'
           # QUESTION: do we really need to evaluate_math in this case?
-          data[key] = to_color evaluate val, data
+          val = evaluate val, data
+          if key == 'table_grid_color' && ::Array === val && val.size == 2 # x,y
+            data[key] = [to_color(val[0]), to_color(val[1])]
```
mojavelinux, line 168: The style in Asciidoctor is to put the round brackets around the method call. So this would be:

```rb
data[key] = [(to_color val[0]), (to_color val[1])]
```

RuboCop doesn't always catch it.

But to make it more clear, I prefer:

```
data[key] = val.map {|e| to_color e }
```

docs/theming-guide.adoc
```diff
@@ -4184,18 +4184,24 @@ The keys in this category control the arrangement and style of tables and table
 (default: $table-border-color)
 |table:
   grid-color: #eeeeee
+table:
+  grid-color: [eeeeee,aaaaaa]
```
mojavelinux, line 4188: Don't add duplicate examples. That will cause confusion. The additional type should be addressed in the second column. So it would read ass:

```
<<colors,Color>> {vbar} <<colors,Color[x,y]>>
```

lib/asciidoctor/pdf/converter.rb
```diff
@@ -2180,10 +2180,23 @@ def convert_table node
           head_border_bottom_width = theme.table_head_border_bottom_width || table_border_width
         end
         [:top, :bottom, :left, :right].each {|edge| border_width[edge] = table_border_width }
+
         table_grid_color = theme.table_grid_color || table_border_color
-        table_grid_style = (theme.table_grid_style || table_border_style).to_sym
+        table_grid_style = theme.table_grid_style || table_border_style
+        if table_grid_style.is_a? Array
```
mojavelinux, line 2186: Please use this notation for type checks:

```rb
::Array === table_grid_style
```

lib/asciidoctor/pdf/converter.rb
```diff
@@ -2180,10 +2180,23 @@ def convert_table node
...
+          raise 'table-grid-style array must contain two styles' if table_grid_style.size != 2
```
mojavelinux, line 2187: I prefer not to do validations in the converter, so this check is not necessary. Let's check for an Array and assume the theme gave us what we expect.

lib/asciidoctor/pdf/converter.rb
```diff
@@ -2180,10 +2180,23 @@ def convert_table node
...
+          table_grid_style = [table_grid_style[0].to_sym, table_grid_style[1].to_sym]
```
mojavelinux, line 2188: I prefer to use a map here:

```rb
table_grid_style = table_grid_style.map(&:to_sym)
```

lib/asciidoctor/pdf/converter.rb
```diff
@@ -2180,10 +2180,23 @@ def convert_table node
...
+        else
+          table_grid_style = Array.new 2, table_grid_style.to_sym
```
mojavelinux, line 2190: This can be simplified to:

```
table_grid_style = [table_grid_style.to_sym]
```

Prawn already understands the single-entry array.

lib/asciidoctor/pdf/converter.rb
```diff
@@ -2180,10 +2180,23 @@ def convert_table node
...
+        end
         table_grid_width = theme.table_grid_width || theme.table_border_width
-        [:cols, :rows].each {|edge| border_width[edge] = table_grid_width }
+        if table_grid_width.is_a? Array
+          raise 'table-grid-width array must contain two numbers separated by comma, e.g [0.5,1]' if table_grid_width.size != 2 || table_grid_width.detect {|x| !x.is_a? Numeric }
```
mojavelinux, line 2194: Same, we don't need the validation here. I prefer to do these sorts of checks in the theme, if we even do them. I defer a lot of them because we expect the theme to be correct. A crash is the right behavior if it is not.

mojavelinux (changes requested):

docs/theming-guide.adoc
Hextremist: That works, but I like `<<colors,Color>> {vbar} [<<colors,x>>,<<colors,y>>]` better.  What do you think?

mojavelinux: It's a nice syntax, but not consistent with what we've done elsewhere in the theming guide. If we want to make that change, it would need to be in a separate issue so we can apply it universally.

lib/asciidoctor/pdf/converter.rb
Hextremist: [kjell@p10s asciidoctor-pdf]$ bin/asciidoctor-pdf -a pdf-theme=theme.yml test.adoc
comparison of String with 0 failed
  Use --trace for backtrace

This is the error I get if I now try to use something not a number in the style.  But I guess you are right that those who make themes can be thought of as experts.  My check only handled the [] case anyway, so it is now removed.

mojavelinux: > I guess you are right that those who make themes can be thought of as experts.

I'm not going to go quite that far. I just don't think the point of usage is the right place for validations because it muddles up the code and is not likely to be comprehensive anyway. It would either be in the theme loader or some sort of theme validator.

Hextremist: So a separate issue for adding sanity checks to the theme_loader.rb then?

docs/theming-guide.adoc
Hextremist: So would you like me to go back to <<colors,Color>> {vbar} <<colors,Color[x,y]>> so we can close this issue?

lib/asciidoctor/pdf/converter.rb
mojavelinux: Yes. Though it will need to be driven by someone in the community because I'm not able to work on something like that right now.

docs/theming-guide.adoc
mojavelinux: Yes please.

mojavelinux:
> In the future, always create an issue for a change request so that we can use the issue tracker and CHANGELOG to track it. 

I also recommend submitting the PR from a branch rather than master. Otherwise, the PR can end up drifting while it's open if you push unrelated changes to master.

mojavelinux (approved):

PR #1896, opened 2021-03-24
resolves #1895 allow theme to control top margin of callout lists

### LINKED ISSUE #1895: Allow theme to control top margin of callout list
When I wrote Japanese text then first line of callout block render on over of source code block. I found this behavior could control `outline_list_item_spacing` key but I didn’t wish to change item spacing.

Conveniently, the soruce code has a task to create a `colist_margin_top` as a todo, so I used it.

irodori-git:


mojavelinux:
I'm going to rename the key, but otherwise accept this change. Thanks!

PR #1899, opened 2021-03-24
Fix #1897

### LINKED ISSUE #1897: Reference to moved source block jumps to wrong page
If a source block doesn't fit on the current page, it is automatically moved to the next page. However, if you click on a reference to the block, you still jump to the old page.

andswitch:
When converting listing blocks, add PDF link destination after `dry_run` phase of `keep_together` to keep the link destination and the block together.

fixes #1897 (I think :smile:)

### DIFF
--- a/lib/asciidoctor/pdf/converter.rb
+++ b/lib/asciidoctor/pdf/converter.rb
@@ -1741,7 +1741,6 @@ module Asciidoctor
 
       # QUESTION: can we avoid arranging fragments multiple times (conums & autofit) by eagerly preparing arranger?
       def convert_listing_or_literal node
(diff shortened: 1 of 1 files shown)


andswitch:
Fix does not work, see https://github.com/asciidoctor/asciidoctor-pdf/issues/1897#issuecomment-806411451

mojavelinux:
The fix is actually not far off. That is the location where we will need to call. But it needs some additional infrastructure which I'm currently adding. I'll document how to do it in the issue. If you submit another PR based on it, then I'll accept it. Though it will also need a test to verify the destination is in the correct spot.

PR #1911, opened 2021-04-09
consistent use of LF line feeds 

ahus1:
….in IDE (via .editorconfig), Git (via .gitattributes) and Rubocop (via .rubocop.yml) to not having false positive message when running on a OS of CRLF linebreaks.

The setting in .gitattributes was already there, so I took that as the leading configuration.

This was something I needed to figure out when developing on Windows. 

I chose to remove some words in the docs, as for me (as a Ruby newbie) first time contributions to the project weren't simple. See https://css-tricks.com/words-avoid-educational-writing/ for similar words. 

(set to "draft" while CI jobs run)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,3 @@
+[*]
+# this matches the setting in .gitattributes
+end_of_line = lf
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -10,6 +10,9 @@ Layout/BeginEndAlignment:
   Enabled: true
 Layout/CaseIndentation:
   EnforcedStyle: end
+Layout/EndOfLine:
+  # this matches the setting in .gitattributes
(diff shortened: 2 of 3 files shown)


mojavelinux:
Looks good to me. Thanks! I'm actually surprised that RuboCop wasn't already enforcing this. So thanks for catching it.

ahus1:
As I understand it: rubocup was enforcing the default line ending on the platform: CRLF on *indows, LF on *inux. Once .gitattributes enforced LF on all platform, RuboCop needed to be configured in the same way. And this is what this PR did.

mojavelinux:
Yep, I get it. I just think RuboCop's default doesn't make much sense in this case. It's a really bad idea for software projects to allow different line endings on different platforms IMO (even if git offers to switch them automatically).

PR #1924, opened 2021-05-03
drop safe_yaml gem in favor of YAML.safe_load in Ruby stdlib

mojavelinux:


mojavelinux:
This may have to wait until we drop Ruby 2.5. However, many major Linux distributions are still shipping Ruby 2.5 in their LTS release streams. So it may be too soon to drop Ruby 2.5.

mojavelinux:
I think if safe_load is not available, we should just map it to regular load. That way, we can move forward without having to break Ruby 2.5 outright.

mojavelinux:
> I think if safe_load is not available, we should just map it to regular load.

The issue isn't about YAML.safe_load being absent, but having different keyword parameters. I added an adapter.

PR #1959, opened 2021-08-18
Fixes typo in gem name

rui-castro:


mojavelinux:
Nice catch! I can't believe I didn't notice this before.

PR #1977, opened 2021-10-04
Fix README reference to uppercase_mb

### LINKED ISSUE #1976: Index customizer broken in 1.6.x?
Had to adjust my docker images today for building stuff, and I noticed that my index customizer script (originally devised in #1174 - now listed in the README) was failing because of not finding `uppercase_mb`:

```
undefined method `uppercase_mb' for #<Asciidoctor::PDF::IndexCatalog:0x0000557242ea0f60>
Did you mean?  uppercase_pcdata
  Use --trace for backtrace
```

From here, it looks like the `upcase` method is probably what we want to use? Looks like that's a standard Ruby string method, so probably wasn't being used before due to not being in the oldest supported ruby?

https://my.diffend.io/gems/asciidoctor-pdf/1.5.4/1.6.0#d2h-121859

if so, the README should be updated.

rpavlik:
This README change could be backported to 1.6.

Fixes #1976

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1128,7 +1128,7 @@ module Asciidoctor::PDF
   IndexCatalog.prepend (::Module.new do
     def store_primary_term name, dest = nil
       store_dest dest if dest
-      category = uppercase_mb (name.delete_prefix 'fn').chr
+      category = (name.delete_prefix 'fn').upcase.chr
       (init_category category).store_term name, dest
     end
   end)


mojavelinux:
Thanks!

mojavelinux:
> This README change could be backported to 1.6.

This section is not in the README on the 1.6.x branch, so no such update is needed.

PR #1984, opened 2021-11-15
resolves #1983 prevent scientific notation ending up in pdf

### LINKED ISSUE #1983: Regression/bug in svg handling from 1.5.4 to 1.6.1 breaks common workflow
### Summary

In asciidoctor-pdf 1.6.1, numbers formatted using scientific notation can end up in the generated pdf and this corrupts the pdf. Some readers can't open them at all (eg Acrobate Reader) whilst others display corrupted content.

### Why this matters

The behavior is unfortunately triggered by a workflow that I suspect is fairly common. As an example, I wouldn't be surprised if this bug is the one causing #1974 

1. Produce an svg image using Inkscape
2. Use Incscape's function to 'Resize document to Selection' to trim the width/height of the svg
3. Include the resulting svg in your asciidoc doc.
4. Use asciidoctor-pdf to generate a pdf

The problem is that Inkscape outputs 'width/height' attributes in the `<svg>` tag that differs from the corresponding viewbox attributes by a very small margin (~1e-5). This small difference trickles through when asciidoctor-pdf/prawn-svg and friends process the svg in such a way as to yield numbers in scientific notation in the pdf document itself and thus render it corrupt.

To add insult to injury, this behavior is dependent on the total magnitude of the width/height as well so it only appears for some combinations of the two.

I've included an asciidoc doc that exposes this bug. It renders ok in v1.5.4 but not in 1.6.1.

[failing.zip](https://github.com/asciidoctor/asciidoctor-pdf/files/7533600/failing.zip)

### Why this happens

This is my conclusion of why the regression is there:

1. prawn 2.4 has increased the precision from 4 to 5 digits in the generated pdf. 
1.1 The change is in the prawn/pdf-core gem, commit `5791488bc5d5469f7acacbdebf0475830a672452`
2. asciidoctor-pdf v1.6.1 has bumped prawn to v2.4 *but includes code   that monkey patches how pdf-core outputs floats.*
3. The monkey patch uses `my_float.to_f.round 5` to get a precision of 5 digits. (the updated, original, pdf-core uses Kernel::format to achieve the same thing).
3.1 Unfortunately, the `round` method in Ruby _starts using scientific notation when the precision > 4 digits_: eg:
3.1.1 `0.00005.round 4 -> 0.0001`
3.1.2. `0.00005.round 5 -> 5.0e-05`
4. The numbers in scientific notation is outputted to the pdf and this apparently make it crash.

In the v2.0.0-dev branch, this have been fixed by removing the monkey patch as part of commit `d2169970727aaf830ca707c4cc94354a28de6ee1`.

### Fixing it

I have a branch based on the 1.6.x branch in my fork that:
 - adds a test that expose the bug
 - removes the monkey patch of pdf-core in the same way as the aforementioned commit `d2169970...`

https://github.com/rillbert/asciidoctor-pdf/tree/rillbert/fix_float_precision_in_pdf

It works, but the problem is that many other tests now fail. I guess it is because they expect a different precision and not because anything is actually broken. In any case, the endeavor of fixing this in my branch is too big for me at the moment and I also suspect that this work have already been done in the 2.0.0 dev branch. 

### Conclusion

We use this (amazing) tool extensively where I work but this makes it impossible for us to use v1.6.1. The only way forward for us is to downgrade to 1.5.4 until this has been fixed.

rillbert:
Fix #1983 by updating the method of rounding floats so that fixed precision is always used for numbers written to the generated pdf file.


mojavelinux:
Thanks!

PR #1991, opened 2022-01-12
resolves #1990 handle uppercase characters in character references

### LINKED ISSUE #1990: Upper-case hex digits in character references are not recognized
Example using &Omega;:

* `&Omega;` does not work (as expected)
* &#937; (`&#937;`) works (as expected)
* &#x003a9 (`&#x003a9`) works (as expected, since https://github.com/asciidoctor/asciidoctor-pdf/issues/486)
* &#x003A9 (`&#x003A9`) does not work (but should)

etihwnad:
Issue #1990 created an extra filtering step in a process to change the case of hex characters.

Though I'm not well up to speed on Ruby, the contribution guidelines were quite clear, hence this PR :)

The commits
* make a failing test,
* adds a test for entities in hrefs (not sure about this one, **please check that this doesn't bork URI behavior**)
* update the parser to detect A-F also

`bundle exec rspec -fd spec/formatted_text_formatter_spec.rb`  reports all passing afterwards.

### DIFF
--- a/lib/asciidoctor/pdf/formatted_text/parser.treetop
+++ b/lib/asciidoctor/pdf/formatted_text/parser.treetop
@@ -113,7 +113,7 @@ grammar Markup
   end
 
   rule character_hex
-    [0-9A-Fa-f] 2..5
+    [0-9a-fA-F] 2..5
   end
 
   rule character_name
--- a/lib/asciidoctor/pdf/formatted_text/parser.treetop
+++ b/lib/asciidoctor/pdf/formatted_text/parser.treetop
@@ -113,7 +113,7 @@ grammar Markup
   end
 
   rule character_hex
-    [0-9a-f] 2..5
+    [0-9A-Fa-f] 2..5
   end
 
   rule character_name
--- a/lib/asciidoctor/pdf/formatted_text/transform.rb
+++ b/lib/asciidoctor/pdf/formatted_text/transform.rb
@@ -7,7 +7,7 @@ module Asciidoctor
         LF = ?\n
         ZeroWidthSpace = ?\u200b
         CharEntityTable = { amp: '&', apos: ?', gt: '>', lt: '<', nbsp: ?\u00a0, quot: '"' }
-        CharRefRx = /&(?:(#{CharEntityTable.keys.join '|'})|#(?:(\d\d\d{0,4})|x([a-f\d][a-f\d][a-f\d]{0,3})));/
+        CharRefRx = /&(?:(#{CharEntityTable.keys.join '|'})|#(?:(\d{2,6})|x(\h{2,5})));/
         HexColorRx = /^#[a-fA-F0-9]{3,6}$/
         TextDecorationTable = { 'underline' => :underline, 'line-through' => :strikethrough }
         ThemeKeyToFragmentProperty = {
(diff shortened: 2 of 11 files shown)


lib/asciidoctor/pdf/formatted_text/parser.treetop
```diff
@@ -113,7 +113,7 @@ grammar Markup
   end
 
   rule character_hex
-    [0-9a-f] 2..5
+    [0-9A-Fa-f] 2..5
```
mojavelinux, line 116: Please put `a-f` before `A-F` (in all cases).

lib/asciidoctor/pdf/formatted_text/transform.rb
```diff
@@ -7,7 +7,7 @@ class Transform
         LF = ?\n
         ZeroWidthSpace = ?\u200b
         CharEntityTable = { amp: '&', apos: ?', gt: '>', lt: '<', nbsp: ?\u00a0, quot: '"' }
-        CharRefRx = /&(?:(#{CharEntityTable.keys.join '|'})|#(?:(\d\d\d{0,4})|x([a-f\d][a-f\d][a-f\d]{0,3})));/
+        CharRefRx = /&(?:(#{CharEntityTable.keys.join '|'})|#(?:(\d{2,6})|x(\h{2,5})));/
```
mojavelinux, line 10: I wrote it as `\d\d\d{0,4}` because it's faster. Please do the same with `\h` (which I'm glad you added!)

spec/formatted_text_formatter_spec.rb
```diff
@@ -210,8 +212,8 @@
       (expect output[0][:link]).to eql 'https://cast.you?v=999999&list=abcde&index=1'
     end
 
-    it 'should decode hexidecimal character references in link href' do
-      output = subject.format '<a href="https://cast.you?v=999999&#x26;list=abcde&#x26;index=1">My Playlist</a>'
+    it 'should decode hexadecimal character references in link href' do
+      output = subject.format '<a href="https://cast.you?v=999999&#x26;list&#x3d;abcde&#x26;index&#x3d;1">My Playlist</a>'
```
mojavelinux, line 216: I don't think this change is needed. The test is already asserting that the hexadecimal character references are decoded.

mojavelinux (changes requested):
This is a great start. I've left some comments to get this ready for merge. Thank you for taking the time to submit this MR!

I want to point out that there are a few other places where a-z is specified without specifying A-Z. If you search the source code (i.e., grep) you will see where they are. Let's make sure we stay consistent.

lib/asciidoctor/pdf/formatted_text/parser.treetop
etihwnad: ASCII ordering dies hard :)

lib/asciidoctor/pdf/formatted_text/transform.rb
etihwnad: Found other places that match hex digits and regularized to this form also.
The speed comment piqued my curiosity and ended with writing up https://agnd.net/blog/ruby-regex-count-benchmark/

etihwnad:
The treetop grammar in `lib/asciidoctor/pdf/formatted_text/parser.treetop` still uses character ranges, but I'm not sure if it makes a difference there.

The last commit https://github.com/asciidoctor/asciidoctor-pdf/pull/1991/commits/7cd085da582db3ad13b2c4339c6a233a8a09c3de tweaks a few characters so digits >9 are actually decoded regardless of case.

Thanks for the thorough review comments!

lib/asciidoctor/pdf/formatted_text/transform.rb
mojavelinux: Wow, I'm really impressed that you did that analysis! That's super awesome. It truly is surprising that unrolling the ranges speeds things up, but observation proves it to be true with this regexp engine.

mojavelinux (approved):

mojavelinux:
Thanks for your hard work on this! You did a great job.

Issue #2004, opened 2022-03-03
Make delimited blocks breakable by default


graphitefriction:
Currently, most delimited blocks are treated as unbreakable in Asciidoctor PDF and there's no way to turn off the setting. (The only exception is the open block, which follows the spec). The unbreakable behavior was applied in order to be able to apply a background and/or border to the block across page boundaries with accuracy. However, once the fix for #2003 in place, this behavior will no longer be mandatory. The new logic can precisely determine a block's extent regardless of where it starts and thus can apply a background and/or border to it. In light of that improvement, we're proposing that the converter return to making breakable the default for delimited blocks and only treating it as unbreakable if requested by the author using the option with the same name.

The fix for #2003 will lay the groundwork for this change. All that needs to be done for this issue is to check for the `unbreakable` option on the block to determine whether "keep together" logic should be activated:

```rb
keep_together = (node.option? 'unbreakable') && !at_page_top?
```

Keep in mind that if a block starts at the top of the page, it's implicitly breakable since it's already given the maximum space possible to avoid a break.

mojavelinux:
It's happening! :scream:

Issue #2016, opened 2022-03-22
Set up kitchen sink test examples


graphitefriction:
To thoroughly exercise the new logic in #2003 and related scenarios, we need a set of examples to test that the extent is correctly computed per block, taking into account block size (shorter than page, taller than page, taller than multiple pages) including anchor and title, page position of block, breakable or unbreakable, nesting level, and indentation. See https://github.com/asciidoctor/asciidoctor-pdf/issues/2003#issuecomment-1063577128 for test cases. The content and code that will satisfy this issue will be bundled with the PR that resolves #2003.

graphitefriction:
The PR for this issue (https://github.com/mojavelinux/asciidoctor-pdf/pull/1) has been submitted for merging into the parent issue branch. I didn't include the resulting PDF of the kitchen sink document in the PR so we can manage the size of the PR until everything is finalized.

PR #2019, opened 2022-04-04
resolves #2003 reimplement logic to arrange block based on extent instead of height

### LINKED ISSUE #2003: Use scratch document to compute extent for content block in primary document
The converter needs to compute a block's boundaries, that is, its extent, so it can determine where to:

* start a block
* draw a background under the block (if applicable)
* draw a border around the block (if applicable)

### Existing logic

Currently, the converter only computes a block's effective height from the top of the page in a scratch document. Since this logic only guarantees that the height is accurate when computed from the top of the page, it requires blocks which don't fit on the current page to be advanced to the next page (i.e., keep together). Still, even when it appears the block can fit on the current page, the margin behavior may be different, resulting in the block being rendered incorrectly. Furthermore, providing only the height of the block makes the boundaries of the background and/or border more complicated to compute, debugging harder, and keep together behavior mandatory.

### New logic

Instead of determining a block's effective height, the converter should compute a block's extent. The extent will always start from the same place in the scratch document as it will start from in the main document.

The extent stores:

* the start page and cursor
* the end page and cursor
* the current page and cursor

The extent also provides helper methods to query the state, such as whether the extent (and hence block) is only a single page or whether it must be advanced to a new page. The new logic should be able to accommodate both breakable and unbreakable blocks.

Here's a summary of the new logic:

* rename `keep_together` DSL method to `arrange_block`
* always set keep together unless cursor is at top of page (changing this behavior will be addressed in a separate issue)
* run `dry_run` with the provided proc to compute the extent of the block in the scratch document
* if `keep_together` is set: convert the block starting from the top of a new page until the end of the page is reached; if the block spills over to another page, start over again from the cursor location in the main document*
* if cursor is at top of page: convert the block fully starting from the top of a new page in the scratch document
* if cursor is not at the top of the page: convert the block starting from the cursor location in the main document until the end of the page is reached; if the block spills over to another page, and the previous page is empty (or only has the block caption), start over again from the top of the page
* compute the extent of the block in the scratch document
* if the block needs to be pushed to a new page, advance the page in the main document
* use the scratch extent to compute the extent in the main document
* run the proc again to convert the content of the block, this time passing in the extent for it to use to draw the background and/or border

To halt conversion of the block early, the converter will raise the `StopIteration` exception provided by Ruby. This allows the dry run process to avoid doing unnecessary work.

(*) Currently, if a block must be split because it doesn't fit on a single page, it's still advanced to the start of the next page. In this situation, the block should start from the cursor. This avoids a big gap at the bottom of the current page when the block must be split anyway.

This issue replaces #789.

mojavelinux:
This PR is the first step towards enabling breakable blocks by default (#2004). It does so by computing the extent of the block in the primary document rather than just its height. That means its able to represent the boundaries of the block before the block is rendered. Furthermore, it uses the scratch document in a much more intelligent way to guarantee that the extent is accurate (with pixel accuracy).

This PR actually contains the code for breakable blocks by default, but only enables it behind a feature flag. That flag will be removed in the PR for #2004.

Here's what's included in this PR:

* replace dry_run with new implementation that computes the extent of the block from the location it will be placed in the primary document (instead of just the block height)
* replace keep_together with arrange block, which can handle rendering both a breakable and unbreakable content block
* use extent to draw border and background under content for block
* move the push/pop scratch calls to arrange block (thus hiding those details from the converter)
* use the new dry_run method to compute the height of an AsciiDoc table cell without rendering any content that will be truncated (stop converting once height of page is exceeded)
* introduce tare content helper in order to keep caption (aka block title) with content block
* shift block to next page if no content in block fits on current page
* use extent object to track extent of TOC


mojavelinux:
Before merging, I think we need to make sure that the CHANGELOG entry captures the whole of what this issue is about.

lib/asciidoctor/pdf/ext/prawn/extensions.rb
```diff
@@ -795,62 +880,169 @@ def scratch?
...
+      # computes the extent of the content. It then returns to the primary document and yields to
+      # the block again, this time passing in the extent of the content. The extent can be used to
+      # draw a border and/or background under the content before inking it.
+      #
+      # This method is intended to enclose the conversion of a single content block, such as a
+      # sidebar or example block. The arrange logic attempts to keep unbreakable content on the same
+      # page, keeps the top caption pinned to the top of the content, computes the extent of the
+      # content for the purpose of drawing a border and/or background underneath it, and ensures
+      # that the extent does not begin near the bottom of a page if the first line of content
+      # doesn't fit. If unbreakable content does not fit on a single page, it is treated as treated
```
graphitefriction, line 902: There's an extra "as" and "treated" in this sentence.

CHANGELOG.adoc
```diff
@@ -69,6 +69,8 @@ Enhancements::
 
 Bug Fixes::
 
+* reimplement arrange block logic by using scratch document to compute extent for content block in primary document (#2003)
```
graphitefriction, line 72: Add 2016 to issue list, i.e. (#2003, #2016).

graphitefriction:
Looks great to me :+1:

mojavelinux:
Let the record show that this PR once had 75 commits ;) I'm going to squash them down to make the git log more comprehensible.

mojavelinux:
And we are merging!

PR #2030, opened 2022-04-15
resolves #483 support role on paragraph to allow theme to control font properties

### LINKED ISSUE #483: Applying custom style to paragraph using a role
I'm searching for a way to apply my custom style to a paragraph, I couldn't see anything in the https://github.com/asciidoctor/asciidoctor-pdf/blob/master/docs/theming-guide.adoc

I would like to define my style, eg.
`mystyle:`
   `font_size: 24`

and apply it to my paragraph:, eg.
`[.mystyle]`
`paragraph content`

Is there a way to achive that in asciidoctor-pdf?

mojavelinux:


mojavelinux:
I'd still like to add support for the `line-height` property.

mojavelinux:
After giving this some more thought, I'm going to pull the line height support out into a separate issue because technical the fix will apply to a lot more than the custom role on a paragraph.

PR #2070, opened 2022-04-24
move custom role usage and creation to new pages

graphitefriction:
- create new page ROOT:roles.adoc to show using a built-in and custom role
- create new page theme:custom-role.adoc to show how to define a custom role
- move Define a role section from theme:role.adoc to theme:custom-role.adoc
- add text-transform key to role category table


graphitefriction:
@mojavelinux If you could review this for technical errors at your convenience, that would be helpful. It's not my greatest work :stuck_out_tongue_winking_eye: but it ticks the boxes enough that we'll be able to link from the What's New page "paragraphs have roles!" and "lead is now a role!" items to these pages.

graphitefriction:
I don't think I put the theme:custom-role.adoc page in the right place in the navigation list. Maybe it should go right above Page Numbers since that page is about customizing the page numbering whereas Fonts is more about what fonts are built in to the themes (though its child pages are about customizing fonts in your theme). Maybe the real problem is I'm not satisfied with where Customize the Theme and Apply a Theme are located in the list :stuck_out_tongue_closed_eyes:

mojavelinux:
I've reviewed and made a few updates. I think this improvement covers the topic well.

> I don't think I put the theme:custom-role.adoc page in the right place in the navigation list.

I like it just below "Customize a Theme". You have to know how to do that to make a custom role...and it's often the first thing you will want to do next. So, to me, it's a natural fit.

mojavelinux (approved):

graphitefriction:
> I like it just below "Customize a Theme".

That seems good to me too. Thanks for the updates @mojavelinux!

mojavelinux:
Thanks for all the improvements to the docs, @graphitefriction!

PR #2081, opened 2022-04-26
add border-style key to decorated block documentation

graphitefriction:
- update border width value type in key tables
- add missing border radius key in key tables

### DIFF
--- a/docs/modules/theme/pages/table.adoc
+++ b/docs/modules/theme/pages/table.adoc
@@ -14,6 +14,7 @@ The keys in the `table` category control the arrangement and style of tables and
 |background-color
 |xref:color.adoc[Color] +
 (default: `transparent`)
+//TODO Where is the transparent value for the table background set in the base theme?
 |[source]
 table:
   background-color: #ffffff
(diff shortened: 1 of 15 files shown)


graphitefriction:
@mojavelinux This PR is solely to do the bare minimum, which is to add the border-style key where it was missing and add the border-width ends, sides array where missing. This PR **isn't** about robustly explaining border-styles and in the cases of some decorated blocks the border-width: [ends,sides]. As we discussed, there needs to be a top-level page about borders - this is where we could also consolidate those little tips that are on several pages that say things like:

> A border is applied to a block image if `border-color` is specified, `border-width` is specified, the value of `border-width` is greater than zero, and the `noborder` role isn't present. The border is drawn above the image on the inside of the box reserved for the image.

and:

> A border is only applied to a button reference if the `border-color` key is specified and the `border-width` key isn't explicitly set to zero.

That page would also be where we do the blanket statement about radius not applying when you specify different styles per side or ends,sides, etc.
**But, that's for a different PR because this one is already biggish, and this is a MVS to unblock What's New.**

Please pay special attention to the changes I made to the admonitions key page and the running content key page, mainly because I'm not sure what a column rule is in these cases and things may be conflated. E. g., *TODO Since column-rule-style follows the pattern of border-style, does that mean column-rule-width follows the pattern of border-width?*

mojavelinux:
> Since column-rule-style follows the pattern of border-style, does that mean column-rule-width follows the pattern of border-width?

Yes. column-rule is effectively a special border which is only vertical and placed in predetermined position.

docs/modules/theme/pages/table.adoc
```diff
@@ -14,6 +14,7 @@ The keys in the `table` category control the arrangement and style of tables and
 |background-color
 |xref:color.adoc[Color] +
 (default: `transparent`)
+//TODO Where is the transparent value for the table background set in the base theme?
```
mojavelinux, line 17: Technically, it's not. It's just effectively transparent. Whenever I write `transparent` as a default background color value, this is usually what I mean. That it's not inherited...it's just not set. But we could just write `not set` here.

mojavelinux:
This looks great to me. It is a lot of keys, and we might discover refinements later. But as far as I can see, it's the right change.

mojavelinux (approved):

graphitefriction:
Great! Thanks @mojavelinux

PR #2096, opened 2022-04-28
resolves #2095 separate align and text-align keys in theme schema

### LINKED ISSUE #2095: Separate align and text-align keys in theme schema
The schema for the theme is currently inconsistent/ambiguous about the name of the key use to align text. Sometimes it's text-align, other times it's align. Yet, align has a slightly different meaning and purpose than text-align. Therefore, these two key names should be cleanly separately.

The purpose of the text-align key is to align the text within a text box (using text alignment semantics). For example, the heading-text-align key can be used to align the text of a heading to the right (currently, the name of the key is heading-align). The heading text box spans the width of the content area. The text-align key controls how the text is aligned within that box.

The purpose of the align key is to align an element horizontally within its container. It's roughly equivalent to the align-self property of an element in a flexbox container (when its direction is row). This property only applies to elements which do not necessarily span the width of content area (image, caption with restricted width, logo, etc).

Update the documentation to recommend the key name `text-align` where appropriate, update the code to use these new key names, and remap theme keys that end with `-align` that should end with `-text-align` for backward compatibility.

mojavelinux:


mojavelinux:
@graphitefriction This PR will need a docs update. Do you want me to go ahead with it?

graphitefriction (approved):
Looks good to me :+1:

mojavelinux:
I've gone ahead with it ;) I also added a new page to describe the value of the align key separate from the text-align key.

PR #2121, opened 2022-05-02
add custom converter use cases page and put custom converters in examples folder

mojavelinux:


graphitefriction:
@mojavelinux I've looked over the new documentation and it looks good to me. I have one question though, what about the "Customize the grouping and sorting rules" section on index-catalog.adoc. Should that be moved into the extend module as a use case?

mojavelinux:
Not that I can see. That's a special kind of override that is patching the primary converter instead of extending it. I think for now we should just leave it there.

PR #2167, opened 2022-05-13
chore: Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>


mojavelinux:
Thank you for bringing this feature of GitHub Actions to my attention. I was not previously aware of it.

After looking through the docs, I see that the preferred value is `read-all` rather than `read`. (See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-assigning-permissions-to-github_token). It can also be set at the top level, which I prefer. Can you update the files accordingly?

I do not want this key to be set on the `release.yml` workflow as that workflow requires use of the token with the write permission to complete the release process. Can you please remove it from that file?

naveensrinivasan:
> Thank you for bringing this feature of GitHub Actions to my attention. I was not previously aware of it.
> 
> After looking through the docs, I see that the preferred value is `read-all` rather than `read`. (See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-assigning-permissions-to-github_token). It can also be set at the top level, which I prefer. Can you update the files accordingly?
> 
> I do not want this key to be set on the `release.yml` workflow as that workflow requires use of the token with the write permission to complete the release process. Can you please remove it from that file?

I took care of them PTAL. Thanks

mojavelinux:
Thank you for making that update.

PR #2172, opened 2022-05-15
Add new theme and fonts

georgeraraujo:
Crimson (serif): https://github.com/skosch/Crimson
Cabin (sans): https://web.archive.org/web/20180303002031/http://www.impallari.com:80/cabin
Free Serif (fallback): http://savannah.gnu.org/projects/freefont/


mojavelinux:
Thanks for this suggestion. However, I do not want to add any more built-in themes to Asciidoctor PDF. Adding a new theme, especially one that introduces fonts, requires a tremendous amount of maintenance and testing over time. We are sticking with Noto as the built-in font family. The theming system gives you all the tools you need to add your own themes, which you can commit to your own repository.

georgeraraujo:
I understand. Thanks.

PR #2178, opened 2022-05-16
add DOM traversal methods to simplify the logic in the converter

mojavelinux:


mojavelinux:
@Mogztter How would you feel about graduating these methods into Asciidoctor core?

Mogztter:
I think that's a good idea. I especially like the `remove` method 👍

PR #2192, opened 2022-05-19
resolves #1933 add support for scale attribute or iw units on pdfwidth attribute on image

### LINKED ISSUE #1933: Scale images in PDF relative to its original size
Now all options to scale images are relative to the width of the content area (https://asciidoctor.org/docs/asciidoctor-pdf/#image-scaling).
I'd like to have an option to scale the image based of it's original size (1300px => 100% of the width, 650px => 50% of the width for example, 325px => 25%, etc)
Prawn seem to be able to do it (https://prawnpdf.org/docs/0.11.1/Prawn/Images.html) with scale argument.

mojavelinux:


mojavelinux:
We need to look into whether we can/should support this for inline images.

mojavelinux:
I've added support for scale on inline images as well.

PR #2201, opened 2022-05-22
resolves #403 advance table to next page if rowspan in first row does not fit in space remaining on current page

### LINKED ISSUE #403: Column spanning multiple (as in many) rows breaks the table layout
**Note**: I really appreciate asciidoctor-pdf, since it does exactly what I need (in most cases) and allows to keep docs for html and pdf aligned easily. So thanks for all the good work!

However, for some documents I work with, the adoc is generated from xml sources, where there tend to be some border cases with excessively long lists of elements and such.

In this specific case I tried to work around the asciidoctor-pdf limitation #6  for putting lists into table cells, by putting the name of  a group into a cell that spans multiple rows and add rows for each group member as follows 

``` asciidoc
[grid="none"]
[cols="<3, <7"]
|=========================
 .10+|Group
 | member 1

 | member 2

 | member 3

 | member 4

 | member 5

 | member 6

 | member 7

 | member 8

 | member 9

 | member 10

|=========================
```

This produces very satisfying PDF output in general:

[tableList10.pdf](https://github.com/asciidoctor/asciidoctor-pdf/files/140657/tableList10.pdf)

However when the Group starts to exceed more than one page, the members colunm (incl. the border) is shown on the last page only.

Here is a cell spanning 100 rows to illustrate the general behavior:

[tableList100.pdf](https://github.com/asciidoctor/asciidoctor-pdf/files/140632/tableList100.pdf)

With the the customized theme I am using where 

``` yaml
page:
  layout: portrait
  margin: [0.9in, 0.7in, 1.3in, 0.7in]
  size: A4
base:
  align: justify
  font_family: Helvetica
  font_size: 10.0
  line_height_length: 12
  line_height: $base_line_height_length / $base_font_size
```

I assume might affect available lines per page, the border line is between 33 and 34 rows:

[tableList34.pdf](https://github.com/asciidoctor/asciidoctor-pdf/files/140637/tableList34.pdf)
[tableList33.pdf](https://github.com/asciidoctor/asciidoctor-pdf/files/140636/tableList33.pdf)

Page breaks seem to be problematic in general for this, since also a 10 row table spanning a page break causes this behavior:

[tableList10time4.pdf](https://github.com/asciidoctor/asciidoctor-pdf/files/140683/tableList10time4.pdf)

You can use the following code (as I did) to generate your own test files (number of rows is first parameter ...)

``` ruby
pre = <<END
[grid="none"]
[cols="<3, <7"]
|=========================
END

File.open("tableList#{ARGV.first}.adoc", "w") do |adocFile|
  adocFile.puts pre ;
  adocFile.puts " .#{ARGV.first}+|Group"
  (1..ARGV.first.to_i).each { |i| adocFile.puts " | member #{i}\n\n"; };
  adocFile.puts '|=========================';
end
```

Let me know, if you need anything else. I'd be happy to help by analyzing this further ... just point me in the right direction ...

mojavelinux:


mojavelinux:
@rillbert FYI, this is the patch you proposed in https://github.com/prawnpdf/prawn-table/issues/106.

PR #2294, opened 2022-07-24
fix(docs): update the download link of NotoSansMonoCJKsc-VF.ttf

IronBlood:
The previous link was for `NotoSansCJKsc-VF.ttf`.


mojavelinux:
Thanks!

PR #2303, opened 2022-08-07
resolves #2297 fix rendering glitches with footnote labels

### LINKED ISSUE #2297: Footnote numbers can get split across lines
If a footnote number falls the end the line, the brackets and numbers can get split across lines. For example:

```
Lorem ipsum dolor sit amet, consectetur adipiscing elit.[
1]
```

This happens because the footnote number is three separate fragments and Prawn has no way to correlate them.

Furthermore, if the footnote number is the last text in the paragraph, it will be placed too high on the line, making it overlap the preceding line.

mojavelinux:
- prevent footnote label from being broken across lines
- keep footnote labe with preceding text if adjacent
- include encoding square brackets in footnote label in link
- remove enclosing square brackets from markers in footnote list


mojavelinux:
I'm concerned this is significant enough that it requires a major release. I'd like to see if we can come up with a way to fix the problem without the style change, then implement the style change in Asciidoctor PDF 3.

mojavelinux:
Replaced by #2312.

PR #2386, opened 2023-01-18
correct example text for padding

jlstrater:
The rest of this page refers to thematic break so I assume this is supposed to be thematic-break instead of sidebar?


mojavelinux:
Thanks Jenn! You're absolutely correct. Good :female_detective:

PR #2393, opened 2023-02-15
Move font-size row down in the quote-cite table

FabienLaunay:
What:
In the quote-cite table, move the font-size row below the font-kerning row.

Why:
Keep consistency with other Theme Keys Reference tables in which keys are ordered alphabetically.

Remark:
Sorry to bother you with this perfectionist pull request ;-) Feel free to ignore it as I guess you must be quite busy developing this great product.


mojavelinux:
My only correction is that you are sending PRs against an inactive branch. The correct branch for docs changes right now is v2.3.x. It's easy for me to fix this PR, but changes will go through quicker in the future if you target the active branch.

PR #2394, opened 2023-02-15
Delete "caption" from the table-caption Key column

FabienLaunay:
What:
In the table-caption table "Key" column, delete the "caption-" characters string.

Why:
Keep consistency with other Theme Keys Reference tables Key columns.

Remark:
Sorry to bother you with this perfectionist pull request ;-) Feel free to ignore it as I guess you must be quite busy developing this great product.


mojavelinux:
Good catch! When these keys were split into different tables, this renaming got missed.

PR #2403, opened 2023-03-12
resolves #2402 add callout-list-marker-font-color theme key to control color of conum marker in callout list

### LINKED ISSUE #2402: Allow color of conums in callout lists to be controlled using the callout-list-marker-font-color theme key
To align with the theme settings for regular lists, allow the color of conum marker in a callout list to be controlled using the `callout-list-marker-font-color` theme key. If this key is not specified, the color of the conum will match the color used for the target conum (the one in the verbatim block).

mojavelinux:


lirantal:
Thanks for landing this!
I assume I need to wait for a new ruby gem release to use this new themable key?

mojavelinux:
That's correct. Though I really only need to look at one other issue first and then I might be able to turn around a patch release soon after.

lirantal:
Awesome. Looking forward to it!

PR #2429, opened 2023-06-13
add section about inline image scaling in table cells

r0ckarong:
As discussed here: https://asciidoctor.zulipchat.com/#narrow/stream/288690-users.2Fasciidoctor-pdf/topic/Inconsistent.20image.20size.20in.20table.20cells/near/365805999

Added clarification to image sizing and small section about inline image behavior in table cells.


mojavelinux:
Ah yes, I forgot to mention that I revised it a bit for accuracy and conciseness. The point remains the same, though.

Issue #1883, opened 2021-02-28
No author in output


rongcuid:
I cannot get authors in the output PDF.

```
$ asciidoctor-pdf --version
Asciidoctor PDF 1.5.4 using Asciidoctor 2.0.12 [https://asciidoctor.org]
Runtime Environment (ruby 2.7.0p0 (2019-12-25 revision 647ee6f091) [x86_64-linux-gnu]) (lc:UTF-8 fs:UTF-8 in:UTF-8 ex:UTF-8)
```

Test file:
```
= Test Document
Author <email@example.org>

Contents
```

mojavelinux:
The authors are only shown when the document has a title page. You can add a title page either by setting the `title-page` attribute in the document header or setting the doctype to book.

rongcuid:
OK, but I think it is also a valid use to have author displayed without a title page. When I have a one-page document, I don't necessarily want a separate title page. Maybe turn this into a feature request?

mojavelinux:
This has already been reported as #988. Feel free to follow-up there. Note, however, that it is not a feature I will be working on, so the implementation will have to come from the community.

das-g:
Based on https://github.com/asciidoctor/asciidoctor-pdf/issues/66#issuecomment-1128562591, https://github.com/asciidoctor/asciidoctor-pdf/issues/988#issuecomment-1128579996 and quite some trail and error, I've come up with

`pdf-converter-article-title-with-author-and-date.rb`
```ruby
class PDFConverterArticleTitleWithAuthorAndDate < (Asciidoctor::Converter.for 'pdf')
  register_for 'pdf'

  def ink_general_heading doc, title, opts
    return super unless opts[:role] == :doctitle
    theme_font :heading_doctitle do
      ink_prose title, align: :center
    end
    revremark = doc.attributes['revremark']
    if doc.author or doc.revdate or revremark
      theme_font :base do
        author_date_separator = (doc.author and doc.revdate) ? " – " : ""
        revremark_separator = ((doc.author or doc.revdate) and revremark) ? " | " : ""
        ink_prose "#{doc.author}#{author_date_separator}#{doc.revdate}#{revremark_separator}#{revremark}", align: :center 
      end
    end
    theme_margin :heading_doctitle, :bottom
  end
end
```
which adds `author`, `revdate` and `revremark` information to articles without a separate title page.

Does this look reasonable? Can it be improved or be made more general (e.g. to support multiple authors)?

[...]

Issue #988, opened 2018-12-31
Document title theming


ogmios-voice:
As a follow up on issues #105 and #135 we should be able to theme the document **title section** in a consistent way while also supporting the requirements of the different document types (i.e. separate title page vs. title section followed by content).

The first question is what should include the **title section**?

* In case of a `book` it is composed of the `logo` (image), `title`, `subtitle`, `authors`, and `revision`.
* In case of an `article` (according to #975) it includes only the `title`.

habamax:
I believe for the article it should include everything but logo (`title`, `subtitle`, `authors` and `revision`)

```
    				 Title
    				Subtitle
    				 Autors
    				Revision

    				 Title
    				Subtitle
    	                Autors           Revision

```

habamax:
And actually, as of now #105 renders Title and Subtitle the same way as for HTML backend: `Title: Subtitle`

ogmios-voice:
Based on comment: https://github.com/asciidoctor/asciidoctor-pdf/issues/105#issuecomment-450627928
I'm suggesting the extension of the theme config with a **new** root level **section** named e.g. `title_block` that could include the same values as the `title_page` section with the exception of the `logo`.

1. **Articles** will use this new `title_block` to theme the `title`, `subtitle`, `authors`, and `revision `fields.
2. **Books** will use the already existing `title_page` configuration.
3. All `title_page` configuration values will **fall back** to the values defined under `title_block`.
4. All `title_block` configuration values will **fall back** to the current fallback values of `title_page`.

What do you think?

mojavelinux:
As I stated in https://github.com/asciidoctor/asciidoctor-pdf/issues/66#issuecomment-1128562591, we've decided to focus on a programmatic solution to this need. Refer to that comment for the reasoning.

If someone would like to come forward with a proposal to change the built-in `ink_title_page` method to look for additional keys, I won't rule it out. However, I think you'll quickly find that no matter how many keys we add, it's always going to feel constrained trying to design through configuration (short of recreating CSS). That's why I think a more powerful approach is to make the converter easier to extend, which is what we've done.

If you think it would be easier if we made an "ink" method for the article title, I'm open to considering that. However, I think you'll find that you can already capture that hook by checking the `opts[:role] == :doctitle` in the `ink_general_heading` method.

```rb
def ink_general_heading doc, title, opts
  return super unless opts[:role] == :doctitle
  theme_font :doctitle do
    ink_prose title, align: :center
[...]

PR #2439, opened 2023-07-17
document how to show author etc. without title page

das-g:
Add a use case example that makes `author`, `revdate` and `revremark` visible in document types (e.g. `article`s) without a title page, by hooking into the converter function for the document title and adding these infos in a line below it.

Workaround for #1883

See https://github.com/asciidoctor/asciidoctor-pdf/issues/1883#issuecomment-1637196810

### DIFF
--- a/docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
+++ b/docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
@@ -4,7 +4,7 @@ class PDFConverterArticleTitleWithAuthorAndDate < (Asciidoctor::Converter.for 'p
   def ink_general_heading doc, title, opts
     return super unless opts[:role] == :doctitle # <1>
     ink_document_title title, opts # <2>
-    ink_document_metainfo doc
+    ink_document_metainfo doc, opts
     ink_titlesection_body_separator
   end
 
@@ -26,14 +26,14 @@ class PDFConverterArticleTitleWithAuthorAndDate < (Asciidoctor::Converter.for 'p
     end
   end
 
-  def ink_document_metainfo doc
+  def ink_document_metainfo doc, opts
     revremark = doc.attr 'revremark' # <3>
     if doc.author || doc.revdate || revremark # <4>
       move_down @theme.heading_h1_documentmetainfo_margin_top || 0
       theme_font_cascade [:base, :heading_h1_documentmetainfo] do
         author_date_separator = (doc.author && doc.revdate) ? ' – ' : '' # <5>
         revremark_separator = ((doc.author || doc.revdate) && revremark) ? ' | ' : '' # <6>
-        ink_prose %(#{doc.author}#{author_date_separator}#{doc.revdate}#{revremark_separator}#{revremark}), align: :center # <7>
+        ink_prose %(#{doc.author}#{author_date_separator}#{doc.revdate}#{revremark_separator}#{revremark}), align: opts[:align] # <7>
       end
     end
   end
--- a/docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
+++ b/docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
@@ -31,8 +31,8 @@ class PDFConverterArticleTitleWithAuthorAndDate < (Asciidoctor::Converter.for 'p
     if doc.author || doc.revdate || revremark # <6>
       move_down @theme.heading_h1_documentmetainfo_margin_top || 0
       theme_font_cascade [:base, :heading_h1_documentmetainfo] do
-        author_date_separator = (doc.author && doc.revdate) ? ' – ' : '' # <7>
-        revremark_separator = ((doc.author || doc.revdate) && revremark) ? ' | ' : '' # <8>
+        author_date_separator = doc.author && doc.revdate ? ' – ' : '' # <7>
+        revremark_separator = (doc.author || doc.revdate) && revremark ? ' | ' : '' # <8>
         ink_prose %(#{doc.author}#{author_date_separator}#{doc.revdate}#{revremark_separator}#{revremark}), align: opts[:align] # <9>
       end
     end
--- a/docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
+++ b/docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
@@ -9,9 +9,9 @@ class PDFConverterArticleTitleWithAuthorAndDate < (Asciidoctor::Converter.for 'p
     revremark = doc.attributes['revremark'] # <3>
     if doc.author || doc.revdate || revremark # <4>
       theme_font :base do
-        author_date_separator = (doc.author && doc.revdate) ? " – " : "" # <5>
-        revremark_separator = ((doc.author || doc.revdate) && revremark) ? " | " : "" # <6>
-        ink_prose "#{doc.author}#{author_date_separator}#{doc.revdate}#{revremark_separator}#{revremark}", align: :center # <7>
+        author_date_separator = (doc.author && doc.revdate) ? ' – ' : '' # <5>
+        revremark_separator = ((doc.author || doc.revdate) && revremark) ? ' | ' : '' # <6>
+        ink_prose %(#{doc.author}#{author_date_separator}#{doc.revdate}#{revremark_separator}#{revremark}), align: :center # <7>
       end
     end
     theme_margin :heading_doctitle, :bottom
--- a/docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
+++ b/docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
@@ -4,7 +4,7 @@ class PDFConverterArticleTitleWithAuthorAndDate < (Asciidoctor::Converter.for 'p
   def ink_general_heading doc, title, opts
     return super unless opts[:role] == :doctitle # <1>
     ink_document_title title, opts # <2>
-    ink_document_metainfo doc, opts # <3>
+    ink_document_details doc, opts # <3>
     margin_bottom @theme[:heading_h1_margin_bottom] || @theme.heading_margin_bottom # <4>
   end
 
@@ -26,7 +26,7 @@ class PDFConverterArticleTitleWithAuthorAndDate < (Asciidoctor::Converter.for 'p
     end
   end
 
-  def ink_document_metainfo doc, opts
+  def ink_document_details doc, opts
     revremark = doc.attr 'revremark' # <5>
     if doc.author || doc.revdate || revremark # <6>
       move_down @theme.heading_h1_details_margin_top || 0 # <7>
--- a/docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
+++ b/docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
@@ -5,7 +5,7 @@ class PDFConverterArticleTitleWithAuthorAndDate < (Asciidoctor::Converter.for 'p
     return super unless opts[:role] == :doctitle # <1>
     ink_document_title title # <2>
     ink_document_metainfo doc
-    theme_margin :heading_doctitle, :bottom
+    ink_titlesection_body_separator
   end
 
   def ink_document_title title
@@ -23,5 +23,9 @@ class PDFConverterArticleTitleWithAuthorAndDate < (Asciidoctor::Converter.for 'p
         ink_prose %(#{doc.author}#{author_date_separator}#{doc.revdate}#{revremark_separator}#{revremark}), align: :center # <7>
       end
     end
+  end
+
+  def ink_titlesection_body_separator
+    theme_margin :heading_doctitle, :bottom
   end
 end
--- a/docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
+++ b/docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
@@ -4,8 +4,8 @@ class PDFConverterArticleTitleWithAuthorAndDate < (Asciidoctor::Converter.for 'p
   def ink_general_heading doc, title, opts
     return super unless opts[:role] == :doctitle # <1>
     ink_document_title title, opts # <2>
-    ink_document_metainfo doc, opts
-    ink_titlesection_body_separator
+    ink_document_metainfo doc, opts # <3>
+    ink_titlesection_body_separator # <4>
   end
 
   def ink_document_title title, opts
@@ -27,13 +27,13 @@ class PDFConverterArticleTitleWithAuthorAndDate < (Asciidoctor::Converter.for 'p
   end
 
   def ink_document_metainfo doc, opts
-    revremark = doc.attr 'revremark' # <3>
-    if doc.author || doc.revdate || revremark # <4>
+    revremark = doc.attr 'revremark' # <5>
+    if doc.author || doc.revdate || revremark # <6>
       move_down @theme.heading_h1_documentmetainfo_margin_top || 0
       theme_font_cascade [:base, :heading_h1_documentmetainfo] do
-        author_date_separator = (doc.author && doc.revdate) ? ' – ' : '' # <5>
-        revremark_separator = ((doc.author || doc.revdate) && revremark) ? ' | ' : '' # <6>
-        ink_prose %(#{doc.author}#{author_date_separator}#{doc.revdate}#{revremark_separator}#{revremark}), align: opts[:align] # <7>
+        author_date_separator = (doc.author && doc.revdate) ? ' – ' : '' # <7>
+        revremark_separator = ((doc.author || doc.revdate) && revremark) ? ' | ' : '' # <8>
+        ink_prose %(#{doc.author}#{author_date_separator}#{doc.revdate}#{revremark_separator}#{revremark}), align: opts[:align] # <9>
       end
     end
   end
--- a/docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
+++ b/docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
@@ -4,6 +4,17 @@ class PDFConverterArticleTitleWithAuthorAndDate < (Asciidoctor::Converter.for 'p
   def ink_general_heading doc, title, opts
     return super unless opts[:role] == :doctitle # <1>
     ink_document_title title # <2>
+    ink_document_metainfo doc
+    theme_margin :heading_doctitle, :bottom
+  end
+
+  def ink_document_title title
+    theme_font :heading_doctitle do
+      ink_prose title, align: :center
+    end
+  end
+
+  def ink_document_metainfo doc
     revremark = doc.attr 'revremark' # <3>
     if doc.author || doc.revdate || revremark # <4>
       theme_font :base do
@@ -11,13 +22,6 @@ class PDFConverterArticleTitleWithAuthorAndDate < (Asciidoctor::Converter.for 'p
         revremark_separator = ((doc.author || doc.revdate) && revremark) ? ' | ' : '' # <6>
         ink_prose %(#{doc.author}#{author_date_separator}#{doc.revdate}#{revremark_separator}#{revremark}), align: :center # <7>
       end
-    end
-    theme_margin :heading_doctitle, :bottom
-  end
-
-  def ink_document_title title
-    theme_font :heading_doctitle do
-      ink_prose title, align: :center
     end
   end
 end
--- a/docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
+++ b/docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
@@ -3,14 +3,26 @@ class PDFConverterArticleTitleWithAuthorAndDate < (Asciidoctor::Converter.for 'p
 
   def ink_general_heading doc, title, opts
     return super unless opts[:role] == :doctitle # <1>
-    ink_document_title title # <2>
+    ink_document_title title, opts # <2>
     ink_document_metainfo doc
     ink_titlesection_body_separator
   end
 
-  def ink_document_title title
-    theme_font :heading_doctitle do
-      ink_prose title, align: :center
+  def ink_document_title title, opts
+    if (top_margin = @theme.heading_h1_margin_page_top || @theme.heading_margin_page_top) > 0
+      move_down top_margin
+    end
+    pad_box @theme.heading_h1_padding do
+      if (transform = resolve_text_transform opts)
+        title = transform_text title, transform
+      end
+      if (inherited = apply_text_decoration font_styles, :heading, 1).empty?
+        inline_format_opts = true
+      else
+        inline_format_opts = [{ inherited: inherited }]
+      end
+      typeset_text_opts = { color: @font_color, inline_format: inline_format_opts }.merge opts
+      typeset_text title, (calc_line_metrics (opts.delete :line_height) || @base_line_height), typeset_text_opts 
     end
   end
 
@@ -26,6 +38,6 @@ class PDFConverterArticleTitleWithAuthorAndDate < (Asciidoctor::Converter.for 'p
   end
 
   def ink_titlesection_body_separator
-    theme_margin :heading_doctitle, :bottom
+    margin_bottom @theme[:heading_h1_margin_bottom] || @theme.heading_margin_bottom
   end
 end
--- a/docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
+++ b/docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
@@ -1,0 +1,19 @@
+class PDFConverterArticleTitleWithAuthorAndDate < (Asciidoctor::Converter.for 'pdf')
+  register_for 'pdf'
+
+  def ink_general_heading doc, title, opts
+    return super unless opts[:role] == :doctitle # <1>
+    theme_font :heading_doctitle do
+      ink_prose title, align: :center # <2>
+    end
+    revremark = doc.attributes['revremark'] # <3>
+    if doc.author or doc.revdate or revremark # <4>
+      theme_font :base do
+        author_date_separator = (doc.author and doc.revdate) ? " – " : "" # <5>
+        revremark_separator = ((doc.author or doc.revdate) and revremark) ? " | " : "" # <6>
+        ink_prose "#{doc.author}#{author_date_separator}#{doc.revdate}#{revremark_separator}#{revremark}", align: :center # <7>
+      end
+    end
+    theme_margin :heading_doctitle, :bottom
+  end
+end
--- a/docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
+++ b/docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
@@ -29,7 +29,8 @@ class PDFConverterArticleTitleWithAuthorAndDate < (Asciidoctor::Converter.for 'p
   def ink_document_metainfo doc
     revremark = doc.attr 'revremark' # <3>
     if doc.author || doc.revdate || revremark # <4>
-      theme_font :base do
+      move_down @theme.heading_h1_documentmetainfo_margin_top || 0
+      theme_font_cascade [:base, :heading_h1_documentmetainfo] do
         author_date_separator = (doc.author && doc.revdate) ? ' – ' : '' # <5>
         revremark_separator = ((doc.author || doc.revdate) && revremark) ? ' | ' : '' # <6>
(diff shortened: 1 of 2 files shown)


das-g:
Is this something to be listed in the changelog?

das-g:
How can I render the asciidoctor-pdf documentation to test this change?

docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
```diff
@@ -0,0 +1,19 @@
...
+    end
+    revremark = doc.attributes['revremark'] # <3>
+    if doc.author or doc.revdate or revremark # <4>
+      theme_font :base do
+        author_date_separator = (doc.author and doc.revdate) ? " – " : "" # <5>
+        revremark_separator = ((doc.author or doc.revdate) and revremark) ? " | " : "" # <6>
+        ink_prose "#{doc.author}#{author_date_separator}#{doc.revdate}#{revremark_separator}#{revremark}", align: :center # <7>
+      end
+    end
+    theme_margin :heading_doctitle, :bottom
```
das-g, line 17: I've taken this line from https://github.com/asciidoctor/asciidoctor-pdf/issues/988#issuecomment-1128579996 and don't really know what it does. I _assume_ it adds some vertical space between the title (or here, the whole head section including the author/date/revremark line) above and the abstract or main content below, but I haven't verified that, yet.

mojavelinux:
In order to test the documentation, you update the [Antora playbook] for docs.asciidoctor.org to point to your local repository instead of the remote one, then specify the branch name in the branches key. For example:

```
  - url: /path/to/asciidoctor-pdf
    branches: show-author-and-date-and-revremark-in-articles-converter-example
    # or use the value HEAD if the branch is currently checked out in git
    start_path: docs
```

Then run Antora:

```
npx antora antora-playbook.yml
```

Alternately, you can create a separate playbook file that only has a single content source that points to your location repository.

mojavelinux:
Please use `||` and `&&` in place of `or` and `and` respectively. The former are the preferred style in the Ruby language.

mojavelinux:
Also, please use single quotes for strings that do not have interpolation (e.g., variable references). For strings that do, use the enclosure `%()` instead of double quotes. The latter is specifically an Asciidoctor style, but one that has worked for us.

mojavelinux:
Here's a more complete implementation:

```rb
class PDFConverterArticleTitleWithAuthorAndDate < (Asciidoctor::Converter.for 'pdf')
  register_for 'pdf'

  def ink_general_heading doc, title, opts
    return super unless opts[:role] == :doctitle
    if (top_margin = @theme.heading_h1_margin_page_top || @theme.heading_margin_page_top) > 0
      move_down top_margin
    end
    pad_box @theme.heading_h1_padding do
      if (transform = resolve_text_transform opts)
        title = transform_text title, transform
      end
      if (inherited = apply_text_decoration font_styles, :heading, 1).empty?
        inline_format_opts = true
      else
        inline_format_opts = [{ inherited: inherited }]
      end
      typeset_text_opts = { color: @font_color, inline_format: inline_format_opts }.merge opts
      typeset_text title, (calc_line_metrics (opts.delete :line_height) || @base_line_height), typeset_text_opts 
    end
    if (revremark = doc.attr 'revremark') || doc.author || doc.revdate
      move_down @theme.heading_h1_details_margin_top || 0
      theme_font_cascade [:base, :heading_h1_details] do
        revdate_sep = doc.author && doc.revdate ? ' &#x2013; ' : ''
        revremark_sep = ((doc.author || doc.revdate) && revremark) ? ' | ' : ''
        ink_prose %(#{doc.author || ''}#{revdate_sep}#{doc.revdate || ''}#{revremark_sep}#{revremark || ''}), align: opts[:align]
      end
    end
    margin_bottom @theme[:heading_h1_margin_bottom] || @theme.heading_margin_bottom
  end
end
```

I have reuse the built-in heading_h1 category since that is the category that targets the doctitle in an article (the only time this method would be called for an h1 heading). I introduced a `details` subcategory to control the formatting of the details line rather than just defaulting to the base font settings.

Obviously, this could be taken much further, but I think it is a good start.

mojavelinux:
Would you be willing to integrate these updates? If not, I'm happy to continue this. I just don't want to step over you if you want to see it through.

das-g:
> Would you be willing to integrate these updates?

In general, yes, but I'll first have to read and understand them, and won't get around to that before the weekend.

mojavelinux:
That's great! I'll give you space to work then. And please don't feel rushed. As you can see from this thread, it even took me months to follow up.

das-g:
> How can I render the asciidoctor-pdf documentation to test this change?

> In order to test the documentation, you update the [Antora playbook] for docs.asciidoctor.org to point to your local repository instead of the remote one, then specify the branch name in the branches key. For example:
> 
> ```
>   - url: /path/to/asciidoctor-pdf
>     branches: show-author-and-date-and-revremark-in-articles-converter-example
>     # or use the value HEAD if the branch is currently checked out in git
>     start_path: docs
> ```
> 
> Then run Antora:
> 
> ```
> npx antora antora-playbook.yml
> ```
> 
> Alternately, you can create a separate playbook file that only has a single content source that points to your location repository.

Huh. You might want to offer a single-command way to do that, that doesn't require the user to clone additional repositories or to modify additional files, or you risk casual contributors (like me :speak_no_evil:) to take on a
The "the most interesting man in the world" meme, captioned "I don't always test my contribution", "but when I do, I do it after getting it merged"
approach. :see_no_evil:

mojavelinux:
This is the strategy we decided to take. I'm happy to add information to the README about how to configure and run the docs on a branch (similar to how we document how to test changes), but we're not going to change the docs set up because someone is unwilling to follow those instructions. If you want to contribute, this is what it takes. I do it and so does everyone else who helps to maintain the docs. If someone doesn't like it, then they are not the right candidate to contribute to this project and there is nothing we can do about that. It is their decision.

mojavelinux:
I have added documentation. As you can see, it is not necessary to clone the docs.asciidoctor.org repository, though you do need to be able to run Antora (either using npx directly or using the Antora container). See https://github.com/asciidoctor/asciidoctor-pdf/blob/main/CONTRIBUTING-CODE.adoc#documentation

das-g:
> I have added documentation. As you can see, it is not necessary to clone the docs.asciidoctor.org repository, though you do need to be able to run Antora (either using npx directly or using the Antora container). See https://github.com/asciidoctor/asciidoctor-pdf/blob/main/CONTRIBUTING-CODE.adoc#documentation

Thank you! That's not single-step-only but (with the new documentation) sufficiently straight forward. :+1:

I wonder whether it could be automated by a script, but that might be hard without having it write to directories that user doesn't expect anything to be written to and it might also be hard to get it sufficiently portable between operating systems.

mojavelinux:
I have considered adding the playbook file for single project mode to the repository, but I'm not convinced I want to encourage that, at least not until Antora Atlas is stable (which supports partial site builds properly). Once it is, I'm willing to revisit my position.

For now, the right way to build all the documentation is through the build at docs.asciidoctor.org. For those contributing, the steps I provided for single project mode should be sufficient. Like I said, if the person is not willing to follow at least those steps, then it doesn't seem to me like it is the right person to be contributing to the docs.

das-g:
In your [more complete implementation](https://github.com/asciidoctor/asciidoctor-pdf/pull/2439#issuecomment-1748391437), @mojavelinux, is purpose of the part
```ruby
    if (top_margin = @theme.heading_h1_margin_page_top || @theme.heading_margin_page_top) > 0
      move_down top_margin
    end
    pad_box @theme.heading_h1_padding do
      if (transform = resolve_text_transform opts)
        title = transform_text title, transform
      end
      if (inherited = apply_text_decoration font_styles, :heading, 1).empty?
        inline_format_opts = true
      else
        inline_format_opts = [{ inherited: inherited }]
      end
      typeset_text_opts = { color: @font_color, inline_format: inline_format_opts }.merge opts
      typeset_text title, (calc_line_metrics (opts.delete :line_height) || @base_line_height), typeset_text_opts 
    end
```
to get the rendering of the document title itself closer to how it's being done by the [default implementation](https://github.com/asciidoctor/asciidoctor-pdf/blob/92348ec3dcb1f7b8bed8c27bba877d657b96ac39/lib/asciidoctor/pdf/converter.rb#L3244-L3294) (with all the settings it might take into account)? If so, does it make sense to imitate or even duplicate that logic here?

Explaining this code for users (that might be unfamiliar with both, Asciidoctor PDF's internal APIs and conventions _and_ the programming language Ruby) in a way that highlights its purpose and enables them to hack on it and modify it to fit their own need seems somewhat challenging.

das-g:
> Explaining this code for users (that might be unfamiliar with both, Asciidoctor PDF's internal APIs and conventions _and_ the programming language Ruby) in a way that highlights its purpose and enables them to hack on it and modify it to fit their own need seems somewhat challenging.

Regarding that, I think we might have the problem, that the section introduced by this PR tries to cater to several potentially very different audiences:
1. users that want to modify the rendering of the document title itself, in a way that fits their own need, but without a need for portability (i.e., not necessarily applicable to other documents) 
2. users that simply want the author, date and/or revision remark to to be visible in a non-book PDF rendering or their document
3. developers that want to create broadly applicable extensions

I believe group 1 (document title modifiers) would need two things:
- a simple example on
  - how to hook into `ink_general_heading` while only having that affect the document title instead of all titles (i.e., the "trick" to delegate to `super` in all other cases)
  - how to render the title in _some_ way (wouldn't need to have a pretty result or to conform to settings the default implementation does take into account)
- a pointer towards the location of the default implementation, so that they can get inspiration there (and copy the whole thing and modify it, or pick and choose individual parts depending on their prowess about Asciidoctor PDF and Ruby)

Group 3 might need an example and some explanations that indicates best practices on what settings, conventions etc. their extension should respect.

Group 2 is tricky. It might contain people unfamiliar with Ruby (and without time to invest in learning it profoundly) that really wouldn't want to get into the Asciidoctor PDF internals. They'd want the document header to be rendered just like the default implementation would do it, respecting all their settings. But they'd also want an example on how to make author, date and/or revision remark visible, simple enough that they can (with their limited knowledge) understand it sufficiently to know what to adapt and what not to modify how the author, date and/or revision remark are shown, without risking to break the rendering of the document title.

Assuming group 2 wants the author, date and/or revision remark to be shown below the document title but above the gap between the document title section and the beginning of the document body (like our example here currently does), maybe a new extension point between the end of the rendering of the actual document title and that gap (thus between the following to code lines:)
https://github.com/asciidoctor/asciidoctor-pdf/blob/92348ec3dcb1f7b8bed8c27bba877d657b96ac39/lib/asciidoctor/pdf/converter.rb#L3291-L3292
and a simple example on how to use that would cater to its need best?

mojavelinux:
> to get the rendering of the document title itself closer to how it's being done by the default implementation 

precisely.

mojavelinux:
> Explaining this code for users (that might be unfamiliar with both, Asciidoctor PDF's internal APIs and conventions and the programming language Ruby) in a way that highlights its purpose and enables them to hack on it and modify it to fit their own need seems somewhat challenging.

I don't really think we need to burden ourselves with explaining it. For the purpose of the use case, it's more important that it is adding the details without breaking the other functionality of the doctitle. If they are really interested in how it works, they can dive into trying to understand it as self study (or by asking in the chat).

mojavelinux:
> users that simply want the author, date and/or revision remark to to be visible in a non-book PDF rendering or their document

This is the target audience and, in my view, sufficient for the purpose of this docs change. It should maintain the existing behavior / configurability while adding the details. Let's not overthink it.

mojavelinux:
> Group 3 might need an example and some explanations that indicates best practices on what settings, conventions etc. their extension should respect.

If people need more explanation, they can ask in the chat. That's what it's there for.

mojavelinux:
> It might contain people unfamiliar with Ruby

We are not and never have catered to this group on this page. This page is for people who know how to extend a converter (or are willing to try) and just want something that will work so they can either use it as is or build on it in time. Again, let's not overthink this.

mojavelinux:
> maybe a new extension point between the end of the rendering of the actual document title and that gap

I don't want to add anything new to Asciidoctor PDF right now. It's necessary to work with what Asciidoctor PDF provides.

das-g:
> > Explaining this code for users (that might be unfamiliar with both, Asciidoctor PDF's internal APIs and conventions and the programming language Ruby) in a way that highlights its purpose and enables them to hack on it and modify it to fit their own need seems somewhat challenging.
> 
> I don't really think we need to burden ourselves with explaining it. For the purpose of the use case, it's more important that it is adding the details without breaking the other functionality of the doctitle. If they are really interested in how it works, they can dive into trying to understand it as self study (or by asking in the chat).

Would it then make sense to (within the example) at least factor it out into a separate method, so that the easily hackable (and thoroughly explained) parts are separate from the touch-this-if-you-know-what-your're-doing parts?

mojavelinux:
> Would it then make sense to (within the example) at least factor it out into a separate method, so that the easily hackable (and thoroughly explained) parts are separate from the touch-this-if-you-know-what-your're-doing parts?

I'd certainly be okay with the override using its own method calls, sure. I tend to use the prefix `ink_` (so it doesn't conflict with other methods that are node handlers). Something like `ink_document_details` or whatever you think is best.

das-g:
I'm not trying to overthink, merely to take a step back and figure out whether we're headed in the right direction. :slightly_smiling_face:

> > users that simply want the author, date and/or revision remark to to be visible in a non-book PDF rendering or their document
> 
> This is the target audience […]

OK. It's the one I originally had in mind with this PR.

> > It might contain people unfamiliar with Ruby
> 
> We are not and never have catered to this group on this page. This page is for people who know how to extend a converter (or are willing to try) and just want something that will work so they can either use it as is or build on it in time. Again, let's not overthink this.

Makes sense.

---

Not all users of Asciidoctor PDF that want a date and author to be shown are (or should be required to be) programmers. (Or am I mistaken about the target audience of Asciidoctor PDF itself?)

So while I agree with both your statements quote above, do they together imply that the target audience of this example isn't really the target audience of the documentation page this PR would currently place it on? :thinking:

(But nonetheless, I guess with dissecting the example into several methods, we have a workable approach. :+1: I'll try to work that out for now.)

das-g:
> > Would it then make sense to (within the example) at least factor it out into a separate method, so that the easily hackable (and thoroughly explained) parts are separate from the touch-this-if-you-know-what-your're-doing parts?
> 
> I'd certainly be okay with the override using its own method calls, sure. I tend to use the prefix `ink_` (so it doesn't conflict with other methods that are node handlers). Something like `ink_document_details` or whatever you think is best.

I was more thinking factoring out the rendering of the document title, but actually, maybe it makes sense to factor out both. I'll try that and see how the example code looks.

docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
```diff
@@ -0,0 +1,43 @@
...
+      typeset_text_opts = { color: @font_color, inline_format: inline_format_opts }.merge opts
+      typeset_text title, (calc_line_metrics (opts.delete :line_height) || @base_line_height), typeset_text_opts 
+    end
+  end
+
+  def ink_document_metainfo doc
+    revremark = doc.attr 'revremark' # <3>
+    if doc.author || doc.revdate || revremark # <4>
+      theme_font :base do
+        author_date_separator = (doc.author && doc.revdate) ? ' – ' : '' # <5>
```
das-g, line 33: Because `&&` has higher precedence than `… ? … : …`, the braces are now redundant. Should they be removed like in https://github.com/asciidoctor/asciidoctor-pdf/pull/2439#issuecomment-1748391437 or be left in for those not aware of Ruby's precedence rules?

das-g, line 33: https://github.com/asciidoctor/asciidoctor-pdf/pull/2439#issuecomment-1748391437 uses `' &#x2013; '` rather than `' – '`. Are non-ASCII characters to be avoided in the source code? I kinda like to see what text snippet in a string literal will kinda "look" like without having to memorize code point numbers, but I can understand that it might be also important to be able to spot non-ASCII characters easily and to be able to distinguish them at sight.

docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
```diff
@@ -0,0 +1,43 @@
...
+        revremark_separator = ((doc.author || doc.revdate) && revremark) ? ' | ' : '' # <6>
```
das-g, line 34: The inner braces (those around `doc.author || doc.revdate`) could be made redundant and removed by replacing the `&&` with the lower-precedence `and` while keeping the `||` as-is. Alternatively, the outer braces could be removed because `&&` has higher precedence than `… ? … : …`. Should any of these steps be made, or should it be left as-is for those not knowing Ruby's precedence rules by heart?

docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
```diff
@@ -0,0 +1,43 @@
...
+        ink_prose %(#{doc.author}#{author_date_separator}#{doc.revdate}#{revremark_separator}#{revremark}), align: :center # <7>
```
das-g, line 35: https://github.com/asciidoctor/asciidoctor-pdf/pull/2439#issuecomment-1748391437 abbreviates "separator" in variable names as `sep`. I've left it spelled out to avoid misinterpretations, but that makes some lines rather long.

mojavelinux:
> I'm not trying to overthink, merely to take a step back and figure out whether we're headed in the right direction.

I certainly appreciate that. I didn't mean for it to sound judgemental, but rather as an invitation to ourselves not to step too far back ;)

mojavelinux:
> Not all users of Asciidoctor PDF that want a date and author to be shown are (or should be required to be) programmers. (Or am I mistaken about the target audience of Asciidoctor PDF itself?)

When it comes to customizing the converter, we can make the assumption there is a programmer (or at least someone doing that task of a programmer) involved. The non-programmer audience must use the software as it is provided, perhaps customizing the theme (though even that requires some programming acumen).

If we were to add this feature into Asciidoctor PDF proper, then it would be available to our whole audience.

docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
mojavelinux: I prefer to use as few braces as possible. So if it works without them, then drop them. That's the rule of thumb I follow.

mojavelinux: > I kinda like to see what text snippet in a string literal will kinda "look" like without having to memorize code point numbers

I prefer not to rely on extended Unicode characters in the source code because it makes it unclear what the character actually is (we can all assume we can recognize ASCII characters).

If you want to make it more clear, you can extract it as a constant as I have done in the converter itself. See https://github.com/asciidoctor/asciidoctor-pdf/blob/main/lib/asciidoctor/pdf/converter.rb#L87 In that case, the preferred syntax is `?\uNNNN`.

docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
mojavelinux: See above.

docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
mojavelinux: I may have done this subconsciously to be honest. I don't really mind it one way or another as long as all uses of the word are the same (i.e., always use sep or separator, but don't mix).

mojavelinux:
> at least someone doing that task of a programmer

I consider anyone willing to edit source code to be a programmer. Whatever level they consider themselves to be is their own determination. I will help them regardless.

docs/modules/extend/pages/use-cases.adoc
```diff
@@ -57,6 +57,37 @@ The method `stroke_horizontal_rule` draws a horizontal line using the specified
...
+----
+include::example$pdf-converter-article-title-with-author-and-date.rb[]
+----
+<.> Fall back to default implementation unless handling the document title.
+<.> Render the document title.
+    See the definition of `ink_document_title` for details.
+<.> Render author, date and revision remark, if they're available.
+    See the definition of `ink_document_metainfo` for details.
+<.> Render a vertical gap between the title section (incl. author, date and revision remark) and the body of the document.
+    See the definition of `ink_titlesection_body_separator` for details.
```
das-g, line 82: Are my callouts for the method invocations useful or do they just add to the length of the callout list? I'd be fine with removing them, if the additional noise and reading length outweigh their contribution to comprehension.

docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
```diff
@@ -0,0 +1,44 @@
...
+      typeset_text_opts = { color: @font_color, inline_format: inline_format_opts }.merge opts
+      typeset_text title, (calc_line_metrics (opts.delete :line_height) || @base_line_height), typeset_text_opts 
+    end
+  end
+
+  def ink_document_metainfo doc, opts
+    revremark = doc.attr 'revremark' # <5>
+    if doc.author || doc.revdate || revremark # <6>
+      move_down @theme.heading_h1_documentmetainfo_margin_top || 0
+      theme_font_cascade [:base, :heading_h1_documentmetainfo] do
```
das-g, line 33: I feel like these two lines could use some callouts explaining what's going on (and maybe some explanation on how to use the theming options they introduce; though maybe better in the prose outside the callouts), but I'm not yet familiar enough with theme keys and how they form categories and sub-categories to phase that confidently. Can you suggest something?

docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
```diff
@@ -0,0 +1,44 @@
...
+        author_date_separator = doc.author && doc.revdate ? %( #{EmDash} ) : '' # <7>
+        revremark_separator = (doc.author || doc.revdate) && revremark ? ' | ' : '' # <8>
+        ink_prose %(#{doc.author}#{author_date_separator}#{doc.revdate}#{revremark_separator}#{revremark}), align: opts[:align] # <9>
```
das-g, line 36: In what case are the ` || ''` fallbacks of https://github.com/asciidoctor/asciidoctor-pdf/pull/2439#issuecomment-1748391437 needed? Are there any cases where these fields hold "falsy" values that aren't empty strings?

mojavelinux: Even though `nil` converts to empty string in Ruby, I tend to use the `|| ''` expression to make it clear I want that behavior. However, for the purpose of this example, it is probably not necessary and can be removed.

docs/modules/extend/pages/use-cases.adoc
mojavelinux: I'm fine with the callout descriptions. My one request is to move the wrapped lines to the left margin as that is the convention we following the docs.

docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
```diff
@@ -0,0 +1,44 @@
+class PDFConverterArticleTitleWithAuthorAndDate < (Asciidoctor::Converter.for 'pdf')
+  register_for 'pdf'
+
+  def ink_general_heading doc, title, opts
+    return super unless opts[:role] == :doctitle # <1>
+    ink_document_title title, opts # <2>
+    ink_document_metainfo doc, opts # <3>
+    ink_titlesection_body_separator # <4>
```
mojavelinux, line 8: I don't think the ink_titlesection_body_separator method is necessary here. I would prefer this logic to be moved back into this method.

docs/modules/extend/pages/use-cases.adoc
```diff
@@ -57,6 +57,46 @@ The method `stroke_horizontal_rule` draws a horizontal line using the specified
...
+if the theme sets a size of that gap.
+Otherwise, fall back to a zero-height "`gap`".
+<.> When we get here, a heading-specific style will already be set.
+(This is ensured by the methods that call `ink_general_heading`.)
+As we don't want author, date and revision remark to look like a heading,
+we first have to reset to the default style using the `:base` category.
+Then, we can apply the `heading_h1_details` category.
+<.> An em-dash to be put between author and date, but only if both aren't empty.
+<.> A vertical bar to be put before the revision remark, but only if it isn't empty and there's actually something else (author, date or both) in front of it.
+<.> Put author, date, revision remark and the separators together using Ruby https://docs.ruby-lang.org/en/3.0/syntax/literals_rdoc.html#label-Strings[string] interpolation and render the resulting string.
```
mojavelinux, line 98: I don't think the link to the Ruby docs is necessary here. First, it's linking to a specific version, which will fall out of date and I will have to update. But I also think that it's up to the reader to find references to the Ruby documentation, because we could be adding links all over the place.

docs/modules/extend/pages/use-cases.adoc
```diff
@@ -57,6 +57,46 @@ The method `stroke_horizontal_rule` draws a horizontal line using the specified
...
+----
+<.> Fall back to default implementation unless handling the document title.
+<.> Render the document title.
+See the definition of `ink_document_title` for details.
+<.> Render author, date and revision remark, if they're available.
+See the definition of `ink_document_metainfo` for details.
+<.> Render a vertical gap between the title section (incl. author, date and revision remark) and the body of the document.
+See the definition of `ink_titlesection_body_separator` for details.
+<.> The revision remark isn't available as a field of `doc`, so we have to get it from the hash `doc.attributes`.
+We could do that with `doc.attributes['revremark']`, but we use the convenience method `doc.attr` instead.
```
mojavelinux, line 84: The information about using `doc.attributes` is overstated. It's also not the preferred way to access attributes anyway, so no one should be using that way.

mojavelinux (changes requested):

mojavelinux:
Can you ensure that the content uses sentence-per-line formatting. That means only creating a newline in a paragraph or principal list item text after a full stop (`.`, `!`, or `?`). There should be no other wrapping in the source text.

docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
```diff
@@ -0,0 +1,40 @@
+class PDFConverterArticleTitleWithAuthorAndDate < (Asciidoctor::Converter.for 'pdf')
+  register_for 'pdf'
+
+  def ink_general_heading doc, title, opts
+    return super unless opts[:role] == :doctitle # <1>
+    ink_document_title title, opts # <2>
+    ink_document_metainfo doc, opts # <3>
```
das-g, line 7: Should `ink_document_metainfo` also be renamed to something with `details`?

mojavelinux: That would be great. Perhaps `ink_document_details`.

das-g: done with 6fd0ed56fd5bc0817c95dbc4a001d141a30d9ed7

docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
das-g: removed redundant braces with f66cef0058baf2e9cd1fc696c2bd4b1dca2070b2

das-g: EmDash is actually already defined there:
https://github.com/asciidoctor/asciidoctor-pdf/blob/5f44ad256d1e7e7ca0ebcb6e1cbf1ee59b5949f8/lib/asciidoctor/pdf/converter.rb#L86-L86
So with 791f0be83d8560ebe8e3b5ff735db0991c6389ca I just use that.

docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
das-g: removed the outer braces with f66cef0058baf2e9cd1fc696c2bd4b1dca2070b2

docs/modules/extend/examples/pdf-converter-article-title-with-author-and-date.rb
das-g: OK. I'm sticking to "separator" everywhere in the example, then.

docs/modules/extend/pages/use-cases.adoc
das-g: I've completely removed mention of `doc.attributes` with bd5563feec8497d4cbda878ce5427d86bb1e629e. Does the resulting callout read alright?

mojavelinux:
Thanks for updating! I'll have a look this weekend.

mojavelinux:
This is excellent. Great work! And thank you for sticking with it and seeing it through.

mojavelinux (approved):

PR #2440, opened 2023-07-25
Remove broken link from import-pdf-pages documentation

r0ckarong:
The link defined here in the documentation points at a defunct website that apparently has been down since at least 2021.

https://web.archive.org/web/20210315000000*/https://fromplantoprototype.com/blog/2019/08/07/importing-pdf-pages-in-asciidoctor-pdf/

The original blog article provided some technical context to the reasoning behind using this feature but from what I can tell it doesn't really add much in terms of usage instructions.


mojavelinux:
Thanks!

PR #2451, opened 2023-09-26
Update title-page-authors key section to fix example source

tobrien:
The source examples in the title-page-authors key section were referencing source that still contained title-page.    I've modified the source examples to align with the name of the key.

### DIFF
--- a/docs/modules/theme/pages/title-page.adoc
+++ b/docs/modules/theme/pages/title-page.adoc
@@ -340,92 +340,81 @@ The keys in the `title-page-authors` category control the display, arrangement a
 |xref:quoted-string.adoc[Quoted AsciiDoc string] +
 (default: `"\{author}"`)
 |[source]
-title-page:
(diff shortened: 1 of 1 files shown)


mojavelinux:
Although still valid, this is not the correct change. Nesting can happen (almost) any time a key contains a hyphen. In other words, the following two are equivalent:

```
title-page:
  authors:
    font-color: #ff0000
```

```
title-page-authors:
  font-color: #ff0000
```

However, the former is preferred since it avoids having to repeat the `title-page-` prefix. In other words, it's a shorthand.

In fact, even this is valid:

```
title:
  page:
    authors:
      font:
        color: #ff0000
```

But this is unnecessarily verbose.

mojavelinux:
I've added a clarification about this to the docs. See https://docs.asciidoctor.org/pdf-converter/latest/theme/keys/#category-keys

PR #2461, opened 2023-11-16
Update fallback-font.adoc

bai-yi-bai:
Added "Default Microsoft Windows Font Limitations" which was discussed in [AsciiDoctor Zulipchat  - How to Find Glyphs to satisfy Font-Fallbacks for CustomTheme](https://asciidoctor.zulipchat.com/#narrow/stream/279727-tips-.F0.9F.92.A1/topic/How.20to.20Find.20Glyphs.20to.20satisfy.20Font-Fallbacks.20for.20CustomTheme).

### DIFF
--- a/docs/modules/theme/pages/fallback-font.adoc
+++ b/docs/modules/theme/pages/fallback-font.adoc
@@ -99,19 +99,24 @@ Now Asciidoctor PDF will look for the emoji in the Symbola font instead of the N
 == Default Microsoft Windows Font Limitations
 When using the fonts provided by default in Microsoft Windows 10, and possibly other versions, it is important to define a fallback font to provide full coverage of Asciidoctor language features.
 In particular, the https://docs.asciidoctor.org/asciidoc/latest/lists/checklist/[Checklist] feature makes use of the _Ballet Box_ ☐ (U+2610) and _Ballot Box with Check_ "☑" (U+2611) characters.
-Two fonts, Segoe UI Symbol (`seuisysm.ttf`) and Segoe UI Emoji (`seguiemj.ttf`), do contain these characters.
+Two fonts provided by Windows, Segoe UI Symbol (`seguisym.ttf`) and Segoe UI Emoji (`seguiemj.ttf`), do contain these characters.
+These two fonts have baselines which match better with most of the Windows font library than the default Notoserif font which is included with Asciidoctor PDF.
 
-Here is an example using Courier with Segoe UI Symbol as a fallback providing the characters needed to support checklists:
+Here is an example using Times New Roman with Segoe UI Symbol as a fallback providing the characters needed to support checklists:
 [,yaml]
 ----
 font:
   catalog:
-    Courier:
-      normal: cour.ttf
-      italic: couri.ttf
-      bold: courbd.ttf
-      bold_italic: courbi.ttf
-    CourierFallback: seuisysm.ttf
+    TimesNewRoman:
+      normal: times.ttf
+      italic: timesi.ttf
+      bold: timesbd.ttf
+      bold_italic: timesbi.ttf
+    TimesNewRomanFallback: seguisym.ttf
   fallbacks:
-  - CourierFallback
+  - TimesNewRomanFallback
 ----
+
+Make sure to configure the theme to include the location of the Windows font files, which is `c:\Windows\Fonts`.
+
+
--- a/docs/modules/theme/pages/fallback-font.adoc
+++ b/docs/modules/theme/pages/fallback-font.adoc
@@ -112,9 +112,9 @@ font:
       italic: timesi.ttf
       bold: timesbd.ttf
       bold_italic: timesbi.ttf
-    TimesNewRomanFallback: seguisym.ttf
+    SegoeUISymbolFallback: seguisym.ttf
   fallbacks:
-  - TimesNewRomanFallback
+  - SegoeUISymbolFallback
 ----
 
 Make sure to configure the theme to include the location of the Windows font files, which is `c:\Windows\Fonts`.
--- a/docs/modules/theme/pages/fallback-font.adoc
+++ b/docs/modules/theme/pages/fallback-font.adoc
@@ -96,10 +96,12 @@ font:
 
 Now Asciidoctor PDF will look for the emoji in the Symbola font instead of the Noto Emoji font.
 
-== Default Microsoft Windows Font Limitations
+== Default Microsoft Windows Font Fallbacks
+
 When using the fonts provided by default in Microsoft Windows 10, and possibly other versions, it is important to define a fallback font to provide full coverage of Asciidoctor language features.
-In particular, the https://docs.asciidoctor.org/asciidoc/latest/lists/checklist/[Checklist] feature makes use of the _Ballet Box_ ☐ (U+2610) and _Ballot Box with Check_ "☑" (U+2611) characters.
-Two fonts provided by Windows, Segoe UI Symbol (`seguisym.ttf`) and Segoe UI Emoji (`seguiemj.ttf`), do contain these characters.
+xref:prepare-custom-font.adoc#modifying-the-font[Modifying the Font] lists all the characters required by Asciidoctor.
+
+Two fonts provided by Windows, Segoe UI Symbol [.path]_seguisym.ttf_ and Segoe UI Emoji [.path]_seguiemj.ttf_, are the only fonts provided by Windows which contain these characters.
 These two fonts have baselines which match better with most of the Windows font library than the default Notoserif font which is included with Asciidoctor PDF.
 
 Here is an example using Times New Roman with Segoe UI Symbol as a fallback providing the characters needed to support checklists:
@@ -117,6 +119,6 @@ font:
   - SegoeUISymbolFallback
 ----
 
-Make sure to configure the theme to include the location of the Windows font files, which is `c:\Windows\Fonts`.
+Make sure to configure the theme to include the location of the Windows font files, which is [.path]_c:\Windows\Fonts_.
 
 
--- a/docs/modules/theme/pages/fallback-font.adoc
+++ b/docs/modules/theme/pages/fallback-font.adoc
@@ -95,3 +95,23 @@ font:
 ----
 
 Now Asciidoctor PDF will look for the emoji in the Symbola font instead of the Noto Emoji font.
+
+== Default Microsoft Windows Font Limitations
+When using the fonts provided by default in Microsoft Windows 10, and possibly other versions, it is important to define a fallback font to provide full coverage of Asciidoctor language features.
+In particular, the https://docs.asciidoctor.org/asciidoc/latest/lists/checklist/[Checklist] feature makes use of the _Ballet Box_ ☐ (U+2610) and _Ballot Box with Check_ "☑" (U+2611) characters.
+Two fonts, Segoe UI Symbol (`seuisysm.ttf`) and Segoe UI Emoji (`seguiemj.ttf`), do contain these characters.
+
+Here is an example using Courier with Segoe UI Symbol as a fallback providing the characters needed to support checklists:
+[,yaml]
+----
+font:
+  catalog:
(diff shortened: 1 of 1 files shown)


docs/modules/theme/pages/fallback-font.adoc
```diff
@@ -95,3 +95,23 @@ font:
 ----
 
 Now Asciidoctor PDF will look for the emoji in the Symbola font instead of the Noto Emoji font.
+
+== Default Microsoft Windows Font Limitations
+When using the fonts provided by default in Microsoft Windows 10, and possibly other versions, it is important to define a fallback font to provide full coverage of Asciidoctor language features.
+In particular, the https://docs.asciidoctor.org/asciidoc/latest/lists/checklist/[Checklist] feature makes use of the _Ballet Box_ ☐ (U+2610) and _Ballot Box with Check_ "☑" (U+2611) characters.
+Two fonts, Segoe UI Symbol (`seuisysm.ttf`) and Segoe UI Emoji (`seguiemj.ttf`), do contain these characters.
+
```
bai-yi-bai, line 103: Sorry, I left out one important part.

"These two fonts have baselines which match better with most of the Windows font library than the default Notoserif font which is included with Asciidoctor PDF."

mojavelinux:
It appears you're suggesting to use Courier as the main font (though it's hard to tell). That's going to make the entire document monospace.  That strikes me as very abnormal. I could see using Courier as the monospace font (for code and codespan), but not the main font. Is there another font from Windows 10 that you would suggest for the main font (for headings and paragraphs)? Keep in mind that the fallback font is used for any text and is not a fallback for a specific font.

It's also not clear from this addition how you are configuring the location of these fonts. Is there a common font directory on Windows 10. If so, the docs should mention it.

I'm also trying to figure out if this change is about using fonts from Windows 10, or it's about configuring a fallback font when using fonts from Windows 10. If it's the former, I think the content belongs on the "Use Custom Fonts Page" instead. That's something I still need to sort out.

docs/modules/theme/pages/fallback-font.adoc
```diff
@@ -95,3 +95,23 @@ font:
 ----
 
 Now Asciidoctor PDF will look for the emoji in the Symbola font instead of the Noto Emoji font.
+
+== Default Microsoft Windows Font Limitations
+When using the fonts provided by default in Microsoft Windows 10, and possibly other versions, it is important to define a fallback font to provide full coverage of Asciidoctor language features.
+In particular, the https://docs.asciidoctor.org/asciidoc/latest/lists/checklist/[Checklist] feature makes use of the _Ballet Box_ ☐ (U+2610) and _Ballot Box with Check_ "☑" (U+2611) characters.
+Two fonts, Segoe UI Symbol (`seuisysm.ttf`) and Segoe UI Emoji (`seguiemj.ttf`), do contain these characters.
```
ioannisstavrakakis, line 102: For Segoe UI Symbol I found `seguisym.ttf` in Windows font folder.

I believe _seuisysm.ttf_ is a typo, please check it out.

bai-yi-bai:
> It appears you're suggesting to use Courier as the main font (though it's hard to tell). That's going to make the entire document monospace. That strikes me as very abnormal. I could see using Courier as the monospace font (for code and codespan), but not the main font. Is there another font from Windows 10 that you would suggest for the main font (for headings and paragraphs)? Keep in mind that the fallback font is used for any text and is not a fallback for a specific font.
> 
> It's also not clear from this addition how you are configuring the location of these fonts. Is there a common font directory on Windows 10. If so, the docs should mention it.
> 
> I'm also trying to figure out if this change is about using fonts from Windows 10, or it's about configuring a fallback font when using fonts from Windows 10. If it's the former, I think the content belongs on the "Use Custom Fonts Page" instead. That's something I still need to sort out.

I wouldn't mind adding more content about Windows-based installs.

Good point about Courier; this was my second choice to use after I found that Calibri doesn't have four separate fonts. A font with four separate files is Times New Roman (times.ttf, timesi.ttf, timesbd.ttf, and timesbi.ttf) 
This also makes me think I ought to also add the courier monospace font to my example.
I'm working off my personal machine(Linux) and looking at my work machine (Windows).

I define the Windows font directory in my antora-assembler.yml's asciidoc, attributes section `pdf-fontsdir: GEM_FONTS_DIR;c:/Windows/Fonts/;`

docs/modules/theme/pages/fallback-font.adoc
bai-yi-bai: Thanks, it is.

docs/modules/theme/pages/fallback-font.adoc
```diff
@@ -95,3 +95,28 @@ font:
...
+[,yaml]
+----
+font:
+  catalog:
+    TimesNewRoman:
+      normal: times.ttf
+      italic: timesi.ttf
+      bold: timesbd.ttf
+      bold_italic: timesbi.ttf
+    TimesNewRomanFallback: seguisym.ttf
```
mojavelinux, line 115: The fallback font should be named after what the font it, not after what it is replacing. In fact, there is no way to say that a fallback font replaces a glyph in a specific font. It's just a fallback for any font when a glyph cannot be found.

mojavelinux:
In regard to my comment:

> It's also not clear from this addition how you are configuring the location of these fonts.

Although you mention it in passing at the end, you don't actually show how it is done. Please say that `pdf-fontsdir` must be configured to include the Windows fonts directory using `-a "pdf-fontsdir=GEM_FONTS_DIR;c:/Windows/Fonts/"` or equivalent. Antora Assembler should not be mentioned here as it's outside the scope of these docs.

mojavelinux:
> I think the content belongs on the "Use Custom Fonts Page" instead. 

I still think this is true. While this advice involves the fallback font, the context is really about how to use Window's built-in fonts as custom fonts...which, in turn, requires the use of the fallback font for the reason cite.

docs/modules/theme/pages/fallback-font.adoc
```diff
@@ -95,3 +95,28 @@ font:
 ----
 
 Now Asciidoctor PDF will look for the emoji in the Symbola font instead of the Noto Emoji font.
+
+== Default Microsoft Windows Font Limitations
+When using the fonts provided by default in Microsoft Windows 10, and possibly other versions, it is important to define a fallback font to provide full coverage of Asciidoctor language features.
```
mojavelinux, line 100: Please include an empty line after any section title you add.

docs/modules/theme/pages/fallback-font.adoc
```diff
@@ -95,3 +95,28 @@ font:
...
+In particular, the https://docs.asciidoctor.org/asciidoc/latest/lists/checklist/[Checklist] feature makes use of the _Ballet Box_ ☐ (U+2610) and _Ballot Box with Check_ "☑" (U+2611) characters.
```
mojavelinux, line 101: Rather than mentioning these two examples, it would be better to point to the following page, which already lists all the characters a custom font is expected to provide. See https://docs.asciidoctor.org/pdf-converter/latest/theme/prepare-custom-font/#required-characters Also, please use a proper xref rather than a qualified URL since the page is within the same site.

docs/modules/theme/pages/fallback-font.adoc
```diff
@@ -95,3 +95,28 @@ font:
...
+      normal: times.ttf
+      italic: timesi.ttf
+      bold: timesbd.ttf
+      bold_italic: timesbi.ttf
+    SegoeUISymbolFallback: seguisym.ttf
+  fallbacks:
+  - SegoeUISymbolFallback
+----
+
+Make sure to configure the theme to include the location of the Windows font files, which is `c:\Windows\Fonts`.
```
mojavelinux, line 120: When referencing paths, please use the following syntax:

```
[.path]_c:/Windows/Fonts_
```

Always use forward slashes in Windows file paths (as it's both legal and avoids backslash escaping issues).

docs/modules/theme/pages/fallback-font.adoc
```diff
@@ -95,3 +95,28 @@ font:
 ----
 
 Now Asciidoctor PDF will look for the emoji in the Symbola font instead of the Noto Emoji font.
+
+== Default Microsoft Windows Font Limitations
+When using the fonts provided by default in Microsoft Windows 10, and possibly other versions, it is important to define a fallback font to provide full coverage of Asciidoctor language features.
+In particular, the https://docs.asciidoctor.org/asciidoc/latest/lists/checklist/[Checklist] feature makes use of the _Ballet Box_ ☐ (U+2610) and _Ballot Box with Check_ "☑" (U+2611) characters.
+Two fonts provided by Windows, Segoe UI Symbol (`seguisym.ttf`) and Segoe UI Emoji (`seguiemj.ttf`), do contain these characters.
```
mojavelinux, line 102: Since these font files are paths, I would prefer the syntax here as well:

```
[.path]_seguisym.ttf_
```

(The stylesheet doesn't include any dedicated styles for the path role at the moment, but it may in the future, at which point it will be able to identify a value that it is a path).

mojavelinux: Thanks for your questions.

As for the page name, I do not want to draw that much attention to using Windows fonts. Therefore, I would like to proceed with adding this new section at the end of the "Use Custom Fonts" page instead.

> I am weary of giving the impression that it's okay to manipulate and repackage fonts provided by Windows due to licensing issues...The scope of this commit is to provide instructions for users using the fonts provided by Windows by default, so that Asciidoctor-PDF files will match their existing OS programs look-and-feel...

As long as fallback fonts are configured that cover all required characters, no modification of the primary font is needed. So the advice would be to keep specifying fallback fonts until all the requirements are met.

Keep in mind that Asciidoctor PDF is not optimized for using Windows Fonts. The application provides a set of built-in fonts that cover all required characters. If a user decided to use their own fonts, they are on their own to find suitable fonts that meet the requirements. Assisting with that is not the responsibility or goal of this project.

> I want to side-step this issue just by referencing what is already installed on a Windows users system.

That does come with some consequences/responsibility as that is not, like I said, a goal of this project.

> I discovered that neither seguisym.ttf and seguiemj.tff contain '\ufeff - zero width no-break space'. Therefore, I need to provide an extra fallback font... what is that character used for?

It looks like this entry is outdated. I can't find any reference to it in the code and rendering seems to work fine without it. I'm going to comment it out as a requirement in the docs for now until we discover otherwise.

> I want to wait for feedback before making any more commits.

Let's not over-complicate this. If you make the changes I requested, I think it will be sufficient to merge.

docs/modules/theme/pages/fallback-font.adoc
bai-yi-bai: Implemented in commit 48c098682b30b0afd6feafaa52614244d4f76c0f.

docs/modules/theme/pages/fallback-font.adoc
bai-yi-bai: Implemented in commit 48c098682b30b0afd6feafaa52614244d4f76c0f.

docs/modules/theme/pages/fallback-font.adoc
bai-yi-bai: I kept it simple, and simply changed the heading in commit 48c098682b30b0afd6feafaa52614244d4f76c0f.

docs/modules/theme/pages/fallback-font.adoc
bai-yi-bai: Implemented in commit 48c098682b30b0afd6feafaa52614244d4f76c0f.

mojavelinux:
Thanks.

Merged as a4bba501135f81fc70df060c8b67427225d9b220. I moved the section to the [Use Custom Fonts](https://docs.asciidoctor.org/pdf-converter/latest/theme/custom-font/) page, did a bit of revision, and added some more references. I believe with the information that's now there, someone should be able to successfully use the system fonts from Microsoft Windows with Asciidoctor PDF.

PR #2462, opened 2023-11-24
Fix small typos in use-cases.adoc

brunchboy:
I think this reads more cleanly but captures the intent of the original text?


mojavelinux:
Thanks! Much appreciated!

PR #2466, opened 2023-11-29
Add disclaimer usage to title page

gxurma:
Hi guys!
Since years I work for companies which need stupid lengthy legal disclaimers on their title page of their documents.
I would like to  add this now
Please revise and use it.
Can be set where on the title page it shall appear.

Also added example in the default theme.

Yours sincerely,
Martin Gyurko
enGYneer.at

### DIFF
--- a/data/themes/default-theme.yml
+++ b/data/themes/default-theme.yml
@@ -114,6 +114,15 @@ title_page:
     font_color: 181818
   revision:
     margin_top: $base_font_size * 1.25
+  disclaimer:
+    top: 80%
+    align: justify
+    font_color: 181818
+    font_size: $base_font_size_small 
+    font_style: normal
+    content: |
+        PROPRIETARY Disclaimer INFORMATION:
+        Whatever you want your Company to disclaim here. Can be a single line text. Some companies need this on their front page so their legal department is happy.
 block:
   margin_bottom: $vertical_rhythm
 caption:


gxurma:
Hi Dan,
Thanks, for pointing me there.
Is there any plan in future to make a similar converter for doing asciidoctor to docx conversion directly? Currently I am doing it with pandoc docbook middleware and that somehow works but is not as good as this pdf generating tool.
Unfortunately many companies still use docx and engineers hate writing in MS word, but their colleagues need the possibility to edit a document further.

best regards,

Martin
enGYneer.at
November 30, 2023 10:08 AM, "Dan Allen" ***@***.*** ***@***.******@***.***>)> wrote:
	We're not going to add an element to the title page to support a one-off use case. However, you can accomplish this using an extended converter. In fact, you can find an example of this exact customization on the use cases page. See https://docs.asciidoctor.org/pdf-converter/latest/extend/use-cases/#custom-title-page (https://docs.asciidoctor.org/pdf-converter/latest/extend/use-cases/#custom-title-page)

	It would look something like this:

	class PDFConverterCustomTitlePage < (Asciidoctor::Converter.for 'pdf') register_for 'pdf' def ink_title_page doc super @y = page_height * 0.2 theme_font :title_page_disclaimer do layout_prose (doc.attr 'disclaimer'), align: ***@***.***_page_disclaimer_align || @theme.title_page_align || :left).to_sym, margin: 0, normalize: false end end end

	There's also an example of how to insert a license/copyright page. See https://docs.asciidoctor.org/pdf-converter/latest/extend/use-cases/#license-page (https://docs.asciidoctor.org/pdf-converter/latest/extend/use-cases/#license-page)

PR #2483, opened 2024-01-29
resolves #2482 add select combining characters to bundled fonts

### LINKED ISSUE #2482: Add select combining characters to bundled fonts
In order to add diacritical marks to certain characters, such as small capital letters, fonts will use combining characters instead of dedicated characters. Add the combining characters in the range U+0300-U+030f for this purpose. (If a larger range is needed, we can add it later).

Here's an example of a letter that uses one of this characters: ᴇ́ (small capital E with an acute accent)

mojavelinux:


mojavelinux:
With this added, it's possible to support the smallcaps transformation (with help from an extended converter) for non-ASCII letters. See https://docs.asciidoctor.org/pdf-converter/latest/theme/text/#transform

PR #2484, opened 2024-01-29
Add clarification about theming to attributes docs

r0ckarong:
According to this discussion, add some clarifications and some minor adjustments to the PDF attributes docs.

I've also added some missing xrefs to the attributes list. I can't really verify if the links work right since I don't know to test the Antora deployment. Sorry :/

https://asciidoctor.zulipchat.com/#narrow/stream/288690-users.2Fasciidoctor-pdf/topic/Self-contained.20adoc.2C.20including.20theme.20configuration.3F

### DIFF
--- a/docs/modules/ROOT/pages/asciidoc-attributes.adoc
+++ b/docs/modules/ROOT/pages/asciidoc-attributes.adoc
@@ -1,12 +1,9 @@
 = AsciiDoc Attributes for PDF
 
 Asciidoctor PDF supports additional built-in document attributes that impact PDF conversion.
-You can use these attributes to control the PDF media, launch mode, and outline, PDF optimization, text hyphenation, and various document-specific theme settings such as the front and back cover images, the background images, the title page logo and background images, page dimensions, and more.
+You can use these attributes to control the PDF media, launch mode, and outline, PDF optimization, text hyphenation, and various document-specific theme settings such as the front and back cover images, the background images, the title page logo and background images, page dimensions, and more. Some of these attributes overlap with settings controlled by the xref:theme:index.adoc[PDF theme].
 
-[NOTE]
-====
-The attributes mentioned on this page allow a limited configuration of the PDF output. For more options over the layout, you will have to xref:theme:create-theme.adoc#extend-default[extend] or create a new xref:theme:index.adoc[PDF theme] and xref:theme:apply-theme.adoc[apply] it to your document during processing.
-====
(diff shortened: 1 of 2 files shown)


docs/modules/ROOT/pages/asciidoc-attributes.adoc
```diff
@@ -138,8 +145,8 @@ If an attribute matches a key in the theme file, the document attribute takes pr
 |Yes
 |`:scripts: cjk`
 
-|`text-align`^[7]^
-|xref:theme:text.adoc#text-align[Text alignment]
+|xref:theme:text.adoc#text-align[text-align] ^[7]^
+|`center` {vbar} `justify` {vbar} `left` {vbar} `right` {vbar} `inherit`
```
mojavelinux, line 149: I don't think this change is needed. When the value is a defined data type (in this case "Text alignment"), let's be consistent and just link to the page that defines the type. This is used throughout the theming documentation.

docs/modules/ROOT/pages/asciidoc-attributes.adoc
```diff
@@ -3,16 +3,23 @@
 Asciidoctor PDF supports additional built-in document attributes that impact PDF conversion.
 You can use these attributes to control the PDF media, launch mode, and outline, PDF optimization, text hyphenation, and various document-specific theme settings such as the front and back cover images, the background images, the title page logo and background images, page dimensions, and more.
 
+[NOTE]
+====
+The attributes mentioned on this page allow a limited configuration of the PDF output. For more options over the layout, you will have to xref:theme:create-theme.adoc#extend-default[extend] or create a new xref:theme:index.adoc[PDF theme] and xref:theme:apply-theme.adoc[apply] it to your document during processing.
+====
```
mojavelinux, line 9: I would prefer this to be written as follows, without an admonition:

> Some of these attributes overlap with settings controlled by the xref:theme:index.adoc[theme]. These attributes are not meant to be a replacement for the theming system. Rather, they're provided as convenience to control some aspects of the theme that are often changed per document or conversion and thus warrant an override. For full control over the rendering of the PDF, you'll want to use a xref:theme:index.adoc[theme], which you can either create anew or xref:theme:create-theme.adoc#extend-default[extend from a built-in theme]. Once you have created a theme, you'll need to xref:theme:apply-theme.adoc[apply it] when converting the document.

r0ckarong: Changed the wording to weave this into the existing wording and removed the admonition.

mojavelinux:
Thanks!

PR #2494, opened 2024-02-14
Fix typo in "Optimize the PDF" preamble

mcp292:


docs/modules/ROOT/pages/optimize-pdf.adoc
```diff
@@ -1,7 +1,7 @@
 = Optimize the PDF
 :url-hexapdf: https://hexapdf.gettalong.org/
 
-By default, Asciidoctor PDF does not optimize the PDF it generates or compresses its streams.
+By default, Asciidoctor PDF does not optimize the PDF it generates, nor compress its streams.
```
mcp292, line 4:
```suggestion
By default, Asciidoctor PDF does not optimize nor compress the PDF it generates.
```

mcp292 (commented):
It could also be simplified to this. Wasn't sure if this loses information.

docs/modules/ROOT/pages/optimize-pdf.adoc
mojavelinux: Technically, the original sentence is correct. Adding nor is a double negative and thus incorrect grammar. However, this is a case where the correct grammar is hard to read. Let's go with the verbose form.

> By default, Asciidoctor PDF does not optimize the PDF it generates and it does not compress its streams.

Technically, optimization is different than compressing streams, which is why they are listed separately.

mojavelinux:
Applied as 65235f20cf03f054a4c7e94c433c22d13f731027

docs/modules/ROOT/pages/optimize-pdf.adoc
mcp292: I should've been more clear about what the "typo" was. I was referring to "compresses" in:

> does not optimize the PDF... or **compresses** its streams

So maybe the original could have been fixed by changing "compresses" to "compress":
```diff
- By default, Asciidoctor PDF does not optimize the PDF it generates or compresses its streams.
+ By default, Asciidoctor PDF does not optimize the PDF it generates or compress its streams.
```

That said, verbose is explicit and explicit is good. I have no objections. Thanks for applying.

mojavelinux: I'll admit I didn't notice the typo fix at first. But when I was looking at the diff when committing it, I finally spotted it. I can't believe that went unnoticed for so long. Typos are sure good at hiding in plain sight.

mcp292: Sure are! Glad to help!

PR #2513, opened 2024-04-02
resolves #2521 don't hyphenate autolink when hyphenation is enabled

### LINKED ISSUE #2521: Don't hyphenate autolink when hyphenation is enabled
If hyphenation is enabled using the hyphens attribute, the URL of an autolink should not be hyphenated. This scenario can be detected by looking for the bare class on the `<a>` tag and skipping over its contents (the URL).

meonkeys:
(see commit log message for details, including caveats)

(sorry about the force push, I won't do any more of those)

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -39,6 +39,7 @@ Improvements::
 * drop support for the unmaintained payment font (`pf`) for use in font-based icons
 * refactor formatted text transform to simplify how inner space is collapsed; verify only inner hard breaks are preserved
 * allow relative font size for sub and sup to be set independently; support combined setting for backwards compatibility
+* don't hyphenate autolink when hyphenation is enabled (#2521) (*@meonkeys*)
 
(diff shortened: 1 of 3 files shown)


mojavelinux:
I've updated the implementation to use a skipping filter instead of a backreference. I think this is a cleaner and more stable approach.

mojavelinux:
Once the tests pass, I will accept this change. A few notes about future PRs that will help them get merged faster.

First, every non-docs PR should have an associated issue. Without an issue, PRs mostly get ignored because they provide any context and are difficult to track using our process. Before submitting a PR, always start by creating an issue. You can mention in the issue that you intend on submitting a PR, which will help get the issue scheduled.

Both the CHANGELOG entry and the PR should reference the issue.

Aside from that, I'd actually say the PR was pretty good. Don't worry about force pushing in a PR. That is completely acceptable. The main goal is that you want to end up with only one commit.

PR #2537, opened 2024-09-08
update to prawn 2.5

noraj:
remove the tmp fix


mojavelinux:
We are not going to upgrade right now. There are unresolved problems with this release and thus we do not consider it stable.

PR #2553, opened 2025-01-07
Update prawn-svg to 0.36.2

### LINKED ISSUE #2551: prawn-svg v0.36.0 dependency update
prawn-svg v0.36.0 is [released](https://github.com/mogest/prawn-svg/releases/tag/v0.36.0). This fixes some issues of svgs being invisible as reported in [Zulip chat](https://asciidoctor.zulipchat.com/#narrow/channel/288690-users.2Fasciidoctor-pdf/topic/Invisible.20svg.20file). Could you update the dependency version?

Fethbita:
Fixes #2551. Updated prawn-svg to 0.36.2.

One failing test was:
```
rspec ./spec/image_spec.rb:1164 # Asciidoctor::PDF::Converter - Image SVG should render linear gradient in SVG
```

But visually there isn't a difference. I added the regenerated file (image-svg-with-gradient.pdf) from under `spec/output`. The `/Creator` field was set to `<feff>` but I changed it to be the same value as the `/Producer` field.

The changes in the error values are expected, see prawn-svg 0.36.1 [release notes](https://github.com/mogest/prawn-svg/releases/tag/v0.36.1).

The change in the gradient handling is expected, see prawn-svg 0.36.0 [release notes](https://github.com/mogest/prawn-svg/releases/tag/v0.36.0):
> The highlight change for this release is a huge rewrite of how gradients work to fix long-standing issues in the implementation as well as adding previously-unsupported properties


mojavelinux:
I'm curious why the error message changed. I'm not sure I understand why that was part of adding more complete gradient support. At least before the user had information about what was wrong with the SVG. I'm not opposed to accepting that the message is more generic, but it does concern me enough to want to follow up on it.

Fethbita:
Created this issue to ask about this https://github.com/mogest/prawn-svg/issues/178

mojavelinux:
Thanks! It looks like that got resolved quickly. I suppose we're just waiting on another release now?

Fethbita:
I'm not sure when the next release will be for prawn-svg and the timeline for asciidoctor-pdf release. So it is up to you to merge this or wait.

I would be happy if this would get merged before the next asciidoctor-pdf release even if prawn-svg is not released. Once prawn-svg is released with the error message fix, it can get updated easily.

mojavelinux:
We cannot merge with a change to the error message, especially given we know it's been fixed upstream. Merging before then would just create unnecessary changes.

Fethbita:
Changed version to 0.36.1 and combined the error messages with the previous error message.

mojavelinux:
This looks good, though I have one minor request. I'd like the `(got 'svg')` to be removed from the assertion. I think that's getting too coupled to the output.

Additionally, `fixture_file 'broken.svg'` needs to be `Regexp.escape fixture_file 'broken.svg'`.

Can you make those updates?

Fethbita:
I have removed the `(got 'svg')` as you requested but adding `Regexp.escape` to `fixture_file 'broken.svg'` breaks the tests on my machine.
```
       expected WARN message containing `~could not embed image in running content: /tmp/asciidoctor-pdf/spec/fixtures/broken.svg; The data supplied is not a valid SVG document.
       Missing end tag for 'rect'' to have been logged, but got WARN: image to embed not found or not readable: /tmp/asciidoctor\-pdf/spec/fixtures/broken\.svg
```

mojavelinux:
You need to use a regular expression, not a string. And you only need to use Regexp.escape if the expected value is a regular expression.

Fethbita:
Ah now I see. Previously, I changed the percent literals `%(~could not ...)` to use forward slash as the separator: `%/~could not .../` since there was also `(got 'svg')` which had parentheses with it. But now since that is deleted, the separators can be turned back to parentheses.

I don't think that regular expression for error checking are needed for these particular tests. Let me know what you think.

mojavelinux:
Yep, that's fine. Looks good.

mojavelinux:
For some reason, the CI checks weren't running on this PR. As a result, we missed that this breaks JRuby 9.2.

PR #2622, opened 2025-11-03
resolves #1415 implement support for extent attribute on tables

### LINKED ISSUE #1415: Support page-wide role for tables
Text content can now be globally indented (theme property `section_indent`, see #737). 
But then, a table that has the `[.stretch]` role is only stretched to that section indent, not to the page margin.
It is useful for some wide tables to stretch them to the page margins.
I suggest the converter should support the role `[.pagewide]` for that. The role should be applicable to tables and to images.

mojavelinux:


mariocup:
Hi @mojavelinux ,
In the file docs/modules/ROOT/pages/table-extent.adoc there is a typo in line 15 **conntainer**

mojavelinux:
Thanks! Fixed.

PR #2634, opened 2026-02-16
fonts: add (Heavy) Ballot X to fallback font

arichardson:
The fallback font already includes \u2713 (CHECK MARK) and \u2714
(HEAVY CHECK MARK), extend this to include (HEAVY) BALLOT X as that is
a common character. For example, \u2718 (Heavy Ballot X) is used by
the RISC-V specification documents.


mojavelinux:
Can we decide between heavy ballot X and heavy multiplication X? It just doesn't seem to me that we need both of them. I agree that having at least one would be an important complement to the checkbox.

arichardson:
> Can we decide between heavy ballot X and heavy multiplication X? It just doesn't seem to me that we need both of them. I agree that having at least one would be an important complement to the checkbox.

Sounds good, I will drop the multiplication.

arichardson:
Updated. I noticed that I also need 0x26AB (Medium Black Circle), can I include that in this pull request or should I submit a follow up?

arichardson:
Ping

mojavelinux:
I'm  busy with other projects at the moment. I will come back to this when that work is complete.

mojavelinux:
To answer your question, no, we are not going to continue adding random glyphs to the fallback font. If there's a companion glyph to a set that is already there, then it's reasonable to add it, as with the original request. If you have other needs, you need to provide your own font.

mojavelinux:
Superseded by #2653.

PR #2638, opened 2026-03-11
Support for cyrillic list numbering

bz007:
Adding lowercyrillic and uppercyrillic numbered list styles, make lowercyrillic/uppercyrillic style default for loweralpha/upperalpha if document lang is 'ru'.


mojavelinux:
Sorry, but this list style is non-standard (not recognized by AsciiDoc or even by CSS) and therefore we cannot add it to this converter. However, you are welcome to extend the converter and add it for your own needs.

Issue #308, opened 2015-08-22
Section id with characters outside of Windows-1252 encoding causes warning


mojavelinux:
If a section id (auto-generated or otherwise) has characters outside of Windows-1252 encoding, a warning is issued. This doesn't seem to break xrefs, but it probably means we need to hex encode them so that they are guaranteed to work in all viewers.

Example:

```
= Document Title

== Проверка

content

<<<

See <<_Проверка>>.
```

mojavelinux:
The solution to this problem is to "hexify" the internal anchors for xrefs. What I can't decide is whether we should do this globally or only if non-ascii characters are detected. I'm leaning towards doing it for all values so it's consistent.

(It would be nice if Prawn did this for us).

flarcher:
I have a similar (same?) issue with a special french character in a section title. But instead of printing a warning, it simply stops the rendering and the PDF gets filled with the _asciidoc_ text (instead of the corresponding PDF content) starting at the point of the section start (where the special character lies).

I am using `org.asciidoctor:asciidoctorj-pdf:1.5.0-alpha.9` (same version as @orloffm in #69) from _Linux Ubuntu_.

mojavelinux:
Can you please provide the sample file. It's best if you are able to serve it over a webserver (your own or Dropbox)? Copy and pasting the content into a web form usually changes the encoding of the text...and hence we are back to square one trying to debug it. Even better is if you can zip it up so as to protect the file from being modified.

flarcher:
Thanks for that fast answer !

Here is the input: 
```
= Hello
blabla
== hehê
hello
image::data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAUAAAAFCAYAAACNbyblAAAAHElEQVQI12P4//8/w38GIAXDIBKE0DHxgljNBAAO9TXL0Y4OHwAAAABJRU5ErkJggg==[]
```
The command is:
```
asciidoctor -B /home/fabrice/tmp -S SAFE -o /home/fabrice/tmp/test.pdf -b pdf -a docdate=2015-09-11 -a data-uri! -a allow-uri-read -a doctime=09:45:04 CEST -a icons=font -
```
And here is the output : https://drive.google.com/file/d/0BwNCc1A1YY17UVlvbU9WclVabUU/view?usp=sharing

flarcher:
[...]

PR #2647, opened 2026-06-24
Fix dead cross-references to discrete headings / inline anchors with non-ASCII IDs

### LINKED ISSUE #2646: Cross-references to `[discrete]` headings and inline anchors with non-ASCII (e.g. Cyrillic) IDs produce dead links
### Asciidoctor version

Asciidoctor PDF 2.3.24 using Asciidoctor 2.0.26 (Ruby 3.4.7, x64-mingw-ucrt, Windows)

### Description

When a `[discrete]` heading — or an inline/phrase anchor (`[[id]]`, `anchor:id[]`, `[#id]#text#`) —
has an **ID containing characters outside Windows-1252** (e.g. Cyrillic), the PDF **named
destination** for that anchor is registered using the **raw UTF-8 bytes** of the ID, while
**cross-references** to it are emitted using the **hex-encoded** form (`0x<utf8-hex>`, the encoding
introduced in #308).

The two names do not match, so every such cross-reference resolves to a non-existent destination —
i.e. a dead link (clicking does nothing; some viewers show it as `…#0x…` / `…#%3A…`).

Regular section titles (`==`, `===`, …) with the **same kind of ID work correctly**, because there
both the destination registration **and** the reference use the hex-encoded form consistently. Only
the **discrete-heading / inline-anchor destination registration** path was missed by the #308 fix.

This is particularly impactful for the **Antora Assembler / `@antora/pdf-extension`**, which converts
every in-page subsection heading into a `[discrete]` heading when it assembles a component into a
single book. In a Russian-language documentation set, that means the *majority* of intra-PDF
cross-references become dead links, even though the same content renders perfect anchors in the HTML
site. 

### Sample document

```asciidoc
= Doc
:doctype: book

== Chapter

See xref:disc_дискрет[discrete link] and xref:sec_секция[section link] and xref:disc_latin[disc latin].

[discrete#disc_дискрет]
==== Дискретный заголовок Cyrillic

Discrete + cyrillic id.

[discrete#disc_latin]
==== Discrete latin id

Discrete + latin id.

[#sec_секция]
== Секция Cyrillic

Real section + cyrillic id.
```

### Expected behavior

All three cross-references resolve to their targets.

### Actual behavior

- `xref:sec_секция[]`   → **works** (real section, Cyrillic id)
- `xref:disc_latin[]`   → **works** (discrete heading, ASCII id)
- `xref:disc_дискрет[]` → **dead link** (discrete heading, Cyrillic id)

### Evidence (registered destinations vs. references in the produced PDF)

**Registered destinations** (`/Names` tree — `(name) N 0 R`):

```
(sec_секция) ...                                       <- real section, Cyrillic : registered in the form that matches refs → OK
(disc_latin) ...                                       <- discrete, ASCII        : OK
(disc_<raw UTF-8 bytes of дискрет>) ...                <- discrete, Cyrillic     : registered as RAW UTF-8 BYTES
                                                          (shows as mojibake "disc_Ð´Ð¸ÑÐºÑÐµÑ" — the UTF-8 bytes read as Latin-1)
```

**Referenced destinations** (link annotations — `/Dest (…)`):

```
(0x646973635fd0b4d0b8d181d0bad180d0b5d182)   <- "disc_дискрет" HEX-encoded → no matching registration → DEAD
(0x7365635f<…секция…>)                        <- "sec_секция"  HEX-encoded → matches section registration → OK
(disc_latin)                                  <- OK
```

So the discrete anchor is **registered** as `disc_<raw-utf8-bytes>` but **referenced** as
`0x<utf8-hex>` → mismatch → dead link. Sections register with the hex form, so they match.

For `disc_дискрет` the hex form is exactly:

```
0x646973635fd0b4d0b8d181d0bad180d0b5d182
   d  i  s  c  _ │д   ││и   ││с   ││к   ││р   ││е   ││т   │
   64 69 73 63 5f d0b4 d0b8 d181 d0ba d180 d0b5 d182
```

## Root cause

- Cross-references encode the destination **name** via `derive_anchor_from_id` (which returns
  `0x<utf8-hex>` for non-ASCII values): see `lib/asciidoctor/pdf/converter.rb`,
  `derive_anchor_from_id` and the `:xref` case (`<a anchor="#{derive_anchor_from_id refid}">`).
- **Sections** register with the derived name too (`convert_section`: `derive_anchor_from_id sect.id`
  then `add_dest_for_block sect, id: sect_anchor`) — so they match.
- But `add_dest_for_block` registers the **raw** id (`add_dest id, node_dest`), and
  `convert_floating_title` (discrete headings) calls it with the raw `node.id`. Inline anchors emit
  `<a id="#{node.id}">` raw (the `:ref` / `:bibref` / `:link` cases and the phrase anchor in
  `convert_inline_quoted`), and that raw id is registered as the destination name by
  `InlineDestinationMarker.render_behind` (`pdf.add_dest name, …`).
- For non-ASCII IDs: registration = raw bytes, reference = `0x<hex>` → mismatch → dead link.

## Related

- #308 — "Section id with characters outside of Windows-1252 encoding causes warning." That issue
  introduced the `0x` hex-encoding for non-Windows-1252 destination names and fixed it for
  sections / references, but the discrete-heading / inline-anchor destination **registration** path
  still writes raw bytes, which is the mismatch reported here.

sandraLbdvTW:
Fixes #2646

## What this fixes

Cross-references to a `[discrete]` heading — or to an inline/phrase anchor (`[[id]]`, `anchor:id[]`,
`[#id]#text#`) — whose **ID contains characters outside Windows-1252** (e.g. Cyrillic) produced
**dead links** in the PDF. References to regular section headings (`==`, `===`) with the same kind of
ID already worked.

The cause is an asymmetry left behind by #308:

- **References** encode the destination name with `derive_anchor_from_id`, which returns
  `0x<utf8-hex>` for any value that is not `ascii_only?` (the `:xref` case uses
  `<a anchor="#{derive_anchor_from_id refid}">`).
- **Sections** register their destination with the *same* derived name (`convert_section` calls
  `derive_anchor_from_id sect.id` and passes the result to `add_dest_for_block`), so they match.
- **Block destinations** (`add_dest_for_block`) and **inline-anchor destinations**
  (`convert_inline_anchor` / `convert_inline_quoted`, registered by
  `InlineDestinationMarker.render_behind`) registered the **raw** id instead.

For a non-ASCII id this means: registration = raw UTF-8 bytes, reference = `0x<utf8-hex>` → the names
don't match → dead link.

## The change

Route destination *registration* through the same `derive_anchor_from_id` that *references* already
use, at the points where the destination name is emitted:

| Location | Path |
| --- | --- |
| `add_dest_for_block` | all block destinations (covers `[discrete]` headings via `convert_floating_title`, and any block with a non-ASCII explicit id) |
| `convert_inline_anchor` — `:link` | inline link with `id=` |
| `convert_inline_anchor` — `:ref` | inline anchor (`[[id]]`, `anchor:id[]`) |
| `convert_inline_anchor` — `:bibref` | bibliography anchor (`[[[id]]]`) |
| `convert_inline_quoted` | phrase anchor (`[#id]#text#`) |

`derive_anchor_from_id` is **idempotent** for values that are already derived or ASCII-only (it returns
them unchanged), so this is a no-op for everything that already worked — sections (already derived),
`toc`, `pdf-anchor`, and every ASCII id. Only non-ASCII block/inline anchor names change, which is
exactly the broken case.

Internal, raw-on-both-sides destination names are deliberately **left untouched** so they keep
matching their (also raw) references: the `_bibref_ref_` citation backlinks, the `_footnoteref_` /
`_footnotedef_` footnote backlinks (integer-suffixed, always ASCII anyway), and index-term anchors
(registered raw and looked up raw by the index).

> Note on scope: the issue describes `[discrete]` headings and `[[id]]` / `anchor:id[]` anchors. The
> fix also covers the phrase anchor `[#id]#text#` (`convert_inline_quoted`), because it shares the
> identical root cause and the existing dest spec already treats `[[details]]` and `[#details]#details#`
> as equivalent inline-anchor forms. If you'd prefer to keep the PR strictly scoped, that one line can
> be dropped — but the bug is present there too.

## Tests

Added two specs in `spec/xref_spec.rb`, next to the existing
`should reference section with ID that contains non-ASCII characters` example:

- `should reference discrete heading with ID that contains non-ASCII characters`
- `should reference inline anchor with ID that contains non-ASCII characters` (covers both
  `[[id]]` and `[#id]#text#` forms)

Each asserts that the registered destination name **and** the link annotation's `/Dest` are both the
hex-encoded id, i.e. they match. Both specs **fail before** this change (the destination is registered
under the raw-bytes key) and **pass after**.

## CHANGELOG

Added an entry under `== Unreleased` → `Bug Fixes::` in `CHANGELOG.adoc`.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -55,6 +55,7 @@ Bug Fixes::
 * prevent special character substitution from interfering with callouts in plain verbatim block (#2390)
 * remove deprecated, undocumented `svg-font-family` theme key (the correct key is `svg-fallback-font-family`)
 * major improvement to OTF support following release of ttfunk 1.8.0 (automatic transitive dependency of prawn)
+* fix dead cross-reference to discrete heading or inline anchor with ID that contains non-ASCII characters by registering destination name using same hex encoding used by reference (#2646)
 
 Compliance::
 
--- a/lib/asciidoctor/pdf/converter.rb
+++ b/lib/asciidoctor/pdf/converter.rb
@@ -2530,7 +2530,7 @@ module Asciidoctor
         target = node.target
         case node.type
         when :link
-          anchor = node.id ? %(<a id="#{node.id}">#{DummyText}</a>) : ''
+          anchor = node.id ? %(<a id="#{derive_anchor_from_id node.id}">#{DummyText}</a>) : ''
           class_attr = ''
           if (role = node.role)
             class_attr = %( class="#{role}")
@@ -2578,14 +2578,14 @@ module Asciidoctor
           end
         when :ref
           # NOTE: destination is created inside callback registered by FormattedTextTransform#build_fragment
-          %(<a id="#{node.id}">#{DummyText}</a>)
+          %(<a id="#{derive_anchor_from_id node.id}">#{DummyText}</a>)
         when :bibref
           id = node.id
           # NOTE: technically node.text should be node.reftext, but subs have already been applied to text
           reftext = (reftext = node.reftext) ? %([#{reftext}]) : %([#{id}])
           reftext = %(<a anchor="_bibref_ref_#{id}">#{reftext}</a>) if @bibref_refs.include? id
           # NOTE: destination is created inside callback registered by FormattedTextTransform#build_fragment
-          %(<a id="#{id}">#{DummyText}</a>#{reftext})
+          %(<a id="#{derive_anchor_from_id id}">#{DummyText}</a>#{reftext})
         else
           log :warn, %(unknown anchor type: #{node.type.inspect})
           nil
@@ -2827,7 +2827,7 @@ module Asciidoctor
         end
 
         # NOTE: destination is created inside callback registered by FormattedTextTransform#build_fragment
-        node.id ? %(<a id="#{node.id}">#{DummyText}</a>#{quoted_text}) : quoted_text
+        node.id ? %(<a id="#{derive_anchor_from_id node.id}">#{DummyText}</a>#{quoted_text}) : quoted_text
       end
 
       # If an id is provided or the node passed as the first argument has an id,
@@ -2851,7 +2851,8 @@ module Asciidoctor
           end
           # TODO: find a way to store only the ref of the destination; look it up when we need it
           node.set_attr 'pdf-destination', (node_dest = (dest_xyz dest_x, dest_y))
-          add_dest id, node_dest
+          # NOTE: derive the destination name the same way references do so non-ASCII IDs match (see #308)
+          add_dest (derive_anchor_from_id id), node_dest
         end
         nil
       end
--- a/spec/xref_spec.rb
+++ b/spec/xref_spec.rb
@@ -104,6 +104,37 @@ describe 'Asciidoctor::PDF::Converter - Xref' do
       (expect (pdf.page 1).text).to include 'See Über Étudier.'
     end
 
+    it 'should reference discrete heading with ID that contains non-ASCII characters' do
+      pdf = to_pdf <<~'END'
+      See <<disc_дискрет>>.
+
+      [discrete#disc_дискрет]
+      == Дискретный заголовок
+      END
+
+      hex_encoded_id = %(0x#{('disc_дискрет'.unpack 'H*')[0]})
+      (expect get_names pdf).to have_key hex_encoded_id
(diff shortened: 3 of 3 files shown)


mojavelinux:
Thanks for getting this started. Now superseded by #2651.

